PR #38, opened 2020-08-12
Cross platform fixture generation + migrate to GHA

wagoodman:
Currently a `pkg/files` test fixture cannot be built on mac since the generation script depends on a GNU tar feature (`--sort`). An easy fix is to containerize this script and extract the fixture tar. When this was done it was apparent that the same circle CI volume-mount-to-host would stop this approach. This this problem is unique to circle CI and would ultimately solve the problem so I elected to move over the checks pipeline to GHA from Circle CI.

Also added the same issue templates as we've done in the other repos.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,20 @@
...
+**What happened**:
+
+**What you expected to happen**:
+
+**How to reproduce it (as minimally and precisely as possible)**:
+
+**Anything else we need to know?**:
+
+**Environment**:
+- grype version (use `grype version -v`):
```
wagoodman, line 19: copy paste problem, I'll remove

PR #39, opened 2020-08-13
Accept image ID as input

### LINKED ISSUE #34: Cannot accept image ID as input
Currently cannot provide a Docker image ID as input (example with grype):
```
go run main.go 34ca5da54570

[ERROR] failed to catalog: could not fetch image '34ca5da54570': unable to trace image save progress: unable to inspect image: Error: No such image: index.docker.io/library/34ca5da54570:latest
exit status 1
%                                                    
```

luhring:
Fixes #34 

This allows us to use local container image IDs in addition to the existing methods of specifying a Docker image. If an image cannot be found locally by specifying the image ID, we fallback to the existing logic for obtaining the image (i.e. pulling).

**Fun fact:** Similar to commands like `docker run`, `docker stop`, and the like, we can now specify images with only a partial image ID. For example...

```
$ syft 6f
...
$ grype a83
...
```

(The above examples require updates to `go.mod` for syft and grype. These PRs coming soon to a theater near you.)

### DIFF
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -16,7 +16,7 @@ var instanceErr error
 var instance *client.Client
 var once sync.Once
 
-func GetClient() (*client.Client, error) {
+func NewClient() (*client.Client, error) {
 	once.Do(func() {
 		var clientOpts = []client.Opt{
 			client.FromEnv,
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -16,7 +16,7 @@ var instanceErr error
 var instance *client.Client
 var once sync.Once
 
-func NewClient() (*client.Client, error) {
+func GetClient() (*client.Client, error) {
 	once.Do(func() {
 		var clientOpts = []client.Opt{
 			client.FromEnv,
--- a/pkg/file/test-fixtures/generators/fixture-1.sh
+++ b/pkg/file/test-fixtures/generators/fixture-1.sh
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -ue
 
+realpath() {
+    [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
+}
+
 FIXTURE_TAR_PATH=$1
 FIXTURE_NAME=$(basename $FIXTURE_TAR_PATH)
 FIXTURE_DIR=$(realpath $(dirname $FIXTURE_TAR_PATH))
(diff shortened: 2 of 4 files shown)


internal/docker/client.go
```diff
@@ -16,7 +16,7 @@ var instanceErr error
 var instance *client.Client
 var once sync.Once
 
-func GetClient() (*client.Client, error) {
```
wagoodman, line 19: nit/question: do we want a convention that distinguishes getting a singleton vs creating a new instance?

luhring: That's a good question. I could see the motivation for `GetClient` as the name since the singleton might be reused rather than creating a new client every time.

My first thought is that the singleton may not be adding much here, and could potentially be removed down the road. For now, I'm happy to revert this name change if necessary.

pkg/file/test-fixtures/generators/fixture-1.sh
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -ue
 
+realpath() {
+    [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
```
alfredodeza, line 5: using `$PWD` would break this depending where this is running from, right? Are we trying to get the absolute path based of the directory where this script exists? or something else?

luhring: I defer to @wagoodman ^

internal/docker/client.go
wagoodman: this was not intended to be a blocking comment, so feel free to leave or change the name. 

I was highlighting the behavior difference compared to a typical constructor and getting a singleton while also asking if want to distinct these things. Agreed on the questionable use of singleton, could probably be removed with no consequence.

luhring: Okay, awesome. I'll revert the name change for now because I think `NewClient` is masking a current reality that a new struct may not actually be created. I'll think more on the singleton topic for later.

pkg/file/test-fixtures/generators/fixture-1.sh
wagoodman: In short, yes. But that is OK.

This is roughly the same behavior as the `realpath` tool:

```
is this an absolute path? if so just return it... otherwise, take the given relative path and resolve it to an absolute path relative to the current working directory
```

So yes, depending on where you run this, the result will be wrong. Given that go will be running this script, it is always relative to the test directory in question it should always give the correct path.

(side note, somewhat unrelated... swapping `$PWD` for `$(pwd)` would be a touch better)

PR #40, opened 2020-09-22
Add OCI support

### LINKED ISSUE #14: Add OCI image support
Currently only supports docker images. The `github.com/google/go-containerregistry` dependency was specifically selected due to the OCI v1 interface abstraction, allowing for multiple providers of container image formats that mapped to OCI v1 features. Since all stereoscope features depend on this abstraction (and are not hard coded to any specific format) it should be a relatively easy lift to implement OCI support from multiple sources (directory, tar archives, etc).

wagoodman:
Adds the ability for users to supply OCI tar archives and OCI directories.

Additionally adds the ability to automatically detect sources (oci archive, oci dir, and docker archive) for a given path.

Closes #14


pkg/file/tarutil.go
```diff
@@ -40,7 +54,7 @@ func ReaderFromTar(reader io.ReadCloser, tarPath string) (io.ReadCloser, error)
 			}, nil
 		}
 	}
-	return nil, fmt.Errorf("file %s not found in tar", tarPath)
+	return nil, ErrFileNotFound
```
luhring, line 57: Just curious — would it be beneficial here to include the file that wasn't found (e.g. from `tarPath`) in the error? (Which would promote `ErrFileNotFound` to its own error type if we want to avoid abusing error strings)

pkg/image/source.go
```diff
@@ -46,22 +64,82 @@ func ParseImageSpec(imageSpec string) (Source, string) {
...
+		return OciTarballSource
 	case "podman":
 		// TODO: implement
 		return UnknownSource
 	}
 	return UnknownSource
 }
 
+// DetectSourceFromPath will distinguish between a oci-layout dir, oci-archive, and a docker-archive.
+func DetectSourceFromPath(imgPath string) (Source, error) {
```
luhring, line 87: It looks like we're never using the `error` return value. If that's true, I'd suggest removing it.

pkg/image/source.go
```diff
@@ -46,22 +64,82 @@ func ParseImageSpec(imageSpec string) (Source, string) {
...
+	return detectSourceFromPath(afero.NewOsFs(), imgPath)
+}
+
+// detectSourceFromPath will distinguish between a oci-layout dir, oci-archive, and a docker-archive for a given filesystem.
+func detectSourceFromPath(fs afero.Fs, imgPath string) (Source, error) {
```
luhring, line 92: It looks like we're never using the `error` return value. If that's true, I'd suggest removing it.

pkg/file/tarutil.go
wagoodman: I had been using this error type in testing, then it was removed, I'll revert the change altogether here.

wagoodman: but to answer your question, yes, this would lean towards an error type in this case. (Would rather not double click on that in this PR)

pkg/image/source.go
```diff
@@ -46,22 +64,82 @@ func ParseImageSpec(imageSpec string) (Source, string) {
...
+		}
+
+		// there are no other directory-based source formats supported
+		return UnknownSource
+	}
+
+	// assume this is an archive...
+	archive, err := fs.Open(imgPath)
+	if err != nil {
+		return UnknownSource
```
alfredodeza, line 113: This pattern is repeated in this function and it somehow doesn't look right, and it worries me that the actual error is being ignored in favor of returning `UnkownSource`. What is the reason to only return a `Source` type here instead of both `Source` and `error` ?

pkg/image/source_test.go
```diff
@@ -102,3 +142,135 @@ func TestParseSource(t *testing.T) {
...
+	return dirPath
+}
+
+func TestDetectSourceFromPath(t *testing.T) {
+	tests := []struct {
+		name           string
+		paths          []string
+		expectedSource Source
+		sourceType     string
+	}{
```
alfredodeza, line 207: it doesn't seem like there is a test case for:

* a path that doesn't exist
* an error produced by `fs.Open()`, I imagine this could be a permissions one? not sure if is feasible
* error from `seek`

pkg/image/source.go
wagoodman: Fair question, I started with including an error and then removed it (https://github.com/anchore/stereoscope/pull/40#discussion_r492986493). When I included it I felt it would make sense to bubble up the error so someone could figure out why detection failed, then I realized I wasn't doing that nor using it. 

This function is meant to be a "best effort" guess as to what the image type is from the given file/dir. But at the same time it's also changing from a "string parser" to additionally a "environment inspector" which is prone to runtime-related errors. I'll head back the direction of returning an error, and bubble it up to the stereoscope client (different from what I was doing earlier).

alfredodeza: Your explanation makes sense to me, I asked the question because it jumped at me, but this could go either way really. Thanks for clarifying!

pkg/file/tarutil.go
wagoodman: jk, due to https://github.com/anchore/stereoscope/pull/40#discussion_r493016378 I'll be adding this back. I'll shoot for a local error type.

pkg/image/source_test.go
wagoodman: for a) there are tests for paths that do not exist, for instance:
```golang
		{
			name:           "oci-layout tar path",
			paths:          []string{"oci-layout"},
			sourceType:     "tar",
			expectedSource: OciTarballSource,
		},
```
this test case will run through checking for a non-existent `manifest.json` file. The `dummy dir paths` and `dummy tar paths` tests will ensure it behaves correctly when all paths will miss.

I've added a test for a non-existent path and changed the behavior around it --given the changed logic I think this will cover the missing fs.Open test (b).

I haven't added a test for a Seek error (c), but feel that the level of testing is probably sufficient.

pkg/image/source.go
wagoodman: updated

pkg/image/source.go
wagoodman: I've added using these errors up to the parent caller

pkg/image/source.go
wagoodman: I've added using these errors up to the parent caller

luhring (approved):
Nice!

PR #49, opened 2020-11-26
Fix handling of opaque directories during file tree merge

### LINKED ISSUE #48: Opaque directories not merged correctly during image squash
Stereoscope is mishandling situations where opaque directories in upper layers are merged down into lower layers, particularly when the directory doesn't exist in the lower layer.

There are two noteworthy problems here:

1. The opaque directory is not added to the lower layer at the end of the merge (and it should be).
1. Because of the above problem, opaque directories are added _implicitly_ during the merge in such a way that prevents these directories from being correctly added to the image's `FileCatalog`, on which consumers depend. Consumers that try to access such directories in a squashed image receive the error "could not find file". This implicit addition of directories during merge operations should be removed.

Tasks:

- [x] add tests for this case
- [x] add the directory from the upper tree even when it contains `.wh..wh..opq` during merge
- [x] don't allow for implicit addition of parents during `AddPath` during merge
- [x] validate that when we create a tree, all its entries are found in the FileCatalog

luhring:
- Fixes #48 
- Adds a unit test to check for opaque directory merging when the directory doesn't exist in the lower tree
- Adds an integration test to check an image squash operation that is known to merge an opaque directory, using a container image based on this bug report: https://github.com/anchore/syft/issues/264
- Amends the file tree `Merge` behavior to _include_ opaque directories from the upper tree in the result tree
- Removes the breadth first walker (it was unused)
- Amends the file tree `AddPath` behavior to require all ancestors to exist for the supplied path, which helps consumers ensure no implicit addition of nodes to the tree when needed. The pre-existing behavior that adds ancestors in addition to the path was moved to `AddPathAndMissingAncestors`, and unit tests are now in place for both variants.
- Adds a check to the layer `Read` operation to ensure that the `FileCatalog` includes an entry for all files in the layer file tree by the time the layer read is considered to be complete


pkg/tree/filetree.go
```diff
@@ -113,36 +122,37 @@ func (t *FileTree) SetFile(f file.Reference) error {
...
-	if err == nil {
-		if pNode, ok := t.pathToFileRef[parent.ID()]; !ok {
-			pNode, err = t.AddPath(parent)
-			if err != nil {
-				return file.Reference{}, err
-			}
-			parentNode = &pNode
-		} else {
-			parentNode = &pNode
+	if path == "/" {
```
wagoodman, line 133: splitting this out into an earlier case made this much more readable 🚀

pkg/tree/breadth_first_walker.go
```diff
@@ -1,79 +0,0 @@
-package tree
```
wagoodman, line 1: this is here as a utility function for callers to work with the tree readers, though it is not used here it's made available for completeness of the API

pkg/tree/filetree.go
```diff
@@ -18,17 +18,26 @@ type FileTree struct {
 
 // NewFileTree creates a new FileTree instance.
 func NewFileTree() *FileTree {
+	tree := newTree()
+
+	// Initialize FileTree with a root "/" node
+	root := file.Path("/")
```
wagoodman, line 24: This seems harmless, but it also makes it impossible to make an empty tree without removing nodes after creating an empty tree. Is this necessary/helpful to change? 

This also seems to counter the need for this line https://github.com/anchore/stereoscope/pull/49/files#diff-f864dd66d754b62339d46b3e43085b83b1d991c4c16730c7932cb52d6a952104R135 .

pkg/tree/filetree.go
```diff
@@ -154,6 +164,35 @@ func (t *FileTree) AddPath(path file.Path) (file.Reference, error) {
 	return f, nil
 }
 
+// AddPathAndMissingAncestors adds a new path to the tree. It also adds any
+// ancestors of the path that are not already present in the tree. The resulting
+// file.Reference of the // new (leaf) addition is returned.
+func (t *FileTree) AddPathAndMissingAncestors(path file.Path) (file.Reference, error) {
```
wagoodman, line 170: nit: maybe consider `AddPathAndAncestors` or `AddPathWithAncestors` instead? The "Missing" is somewhat redundant since `AddPath` only adds paths that have not already been added, so in similar fashion adding a path and ancestors doesn't need an indication of missing-or-not in the name. (Also makes the name shorter without losing meaning, which is an API win)

wagoodman (commented):
Overall solid fix! Only had a couple of relatively small comments to chat through.

pkg/tree/filetree.go
luhring: Discussed — we think it's okay to assert that FileTree always has a root path (`/`), but the `RemovePath` logic needs to be amended so as not to allow the removal of this root path.

wagoodman (approved):

PR #51, opened 2020-12-03
Remove assertion that all files in the FileTree exist in the FileCatalog

luhring:
While discussing the resolution for https://github.com/anchore/stereoscope/issues/48, it was decided that we should check to make sure that during a layer read, all files in a layer's FileTree have corresponding entries in the FileCatalog. This was because previously, tree merges had been adding files to source trees on the fly if they were missing.

This PR removes the assertion that all files in the tree exist in the FileCatalog.

This case was surfaced in https://github.com/anchore/syft/issues/284.

### DIFF
--- a/pkg/image/layer.go
+++ b/pkg/image/layer.go
@@ -1,8 +1,6 @@
 package image
 
 import (
-	"errors"
-
 	"github.com/anchore/stereoscope/internal/bus"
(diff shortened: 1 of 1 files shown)


wagoodman (changes requested):
The file catalog is intended to only capture files that positively exist in the image (in the tar headers)

wagoodman (approved):

PR #52, opened 2020-12-10
Add raw config image option + enhance tags

wagoodman:
- Adds the `WithConfig` option to attach an image config to the metadata
- Enhances the sourcing of tags from the docker daemon to include `RepoTags` from a `docker inspect` call (API, not CLI) for cases when only the ID or partial ID is given (in which case no tags would be included unless explicitly retrieved from an inspect call).

### DIFF
--- a/integration/fixture_image_simple_test.go
+++ b/integration/fixture_image_simple_test.go
@@ -34,7 +34,9 @@ func TestSimpleImage(t *testing.T) {
 			source:         "docker",
 			imageMediaType: v1Types.DockerManifestSchema2,
 			layerMediaType: v1Types.DockerLayer,
-			tagCount:       1,
+			// name:hash
+			// name:latest
+			tagCount: 2,
 		},
 		{
 			name:           "FromOciTarball",
@@ -78,7 +80,7 @@ func assertImageSimpleMetadata(t *testing.T, i *image.Image, expectedValues test
 		t.Errorf("unexpected image media type: %+v", i.Metadata.MediaType)
 	}
 	if len(i.Metadata.Tags) != expectedValues.tagCount {
-		t.Errorf("unexpected number of tags: %d", len(i.Metadata.Tags))
+		t.Errorf("unexpected number of tags: %d : %+v", len(i.Metadata.Tags), i.Metadata.Tags)
 	} else if expectedValues.tagCount > 0 {
 		if !strings.HasPrefix(i.Metadata.Tags[0].String(), fmt.Sprintf("%s-image-simple:", imagetest.ImagePrefix)) {
 			t.Errorf("unexpected image tag: %+v", i.Metadata.Tags)
--- a/internal/stringset.go
+++ b/internal/stringset.go
@@ -1,5 +1,7 @@
 package internal
 
+import "sort"
+
 type Set map[string]struct{}
 
 func NewStringSet() Set {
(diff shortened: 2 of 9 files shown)


integration/fixture_image_simple_test.go
```diff
@@ -78,7 +80,7 @@ func assertImageSimpleMetadata(t *testing.T, i *image.Image, expectedValues test
 		t.Errorf("unexpected image media type: %+v", i.Metadata.MediaType)
 	}
 	if len(i.Metadata.Tags) != expectedValues.tagCount {
-		t.Errorf("unexpected number of tags: %d", len(i.Metadata.Tags))
+		t.Errorf("unexpected number of tags: %d : %+v", len(i.Metadata.Tags), i.Metadata.Tags)
```
luhring, line 83: Very helpful!

internal/stringset.go
```diff
@@ -18,3 +20,15 @@ func (s Set) Contains(i string) bool {
 	_, ok := s[i]
 	return ok
 }
+
+// ToSlice returns a sorted slice of strings that are contained within the set.
+func (s Set) ToSlice() []string {
```
luhring, line 25: khan

pkg/image/docker/manifest.go
```diff
@@ -70,10 +70,10 @@ func extractManifest(tarPath string) (*dockerManifest, error) {
 }
 
 // generateOCIManifest takes a docker manifest and a path to the tar and generates an OCI manifest derived from the given arguments and the docker config.
-func generateOCIManifest(tarPath string, manifest *dockerManifest) (*v1.Manifest, error) {
+func generateOCIManifest(tarPath string, manifest *dockerManifest) (*v1.Manifest, []byte, error) {
```
luhring, line 73: We should update the doc comment to make it clear what the []byte return value is

pkg/image/image.go
```diff
@@ -59,18 +61,22 @@ func WithManifestDigest(digest string) AdditionalMetadata {
...
+	return func(image *Image) error {
+		image.Metadata.RawConfig = config
+		image.Metadata.ID = fmt.Sprintf("sha256:%x", sha256.Sum256(config))
+		return nil
+	}
+}
+
 // NewImage provides a new, unread image object.
-func NewImage(image v1.Image, additionalMetadata ...AdditionalMetadata) (*Image, error) {
+func NewImage(image v1.Image, additionalMetadata ...AdditionalMetadata) *Image {
```
luhring, line 73: No need to rehash this right now, but this API is getting a bit less intuitive (and at the same time, I don't have a strong "better" suggestion yet!). It looks like we're not using these input functions here, and we only use them on `Read`... I don't think this is obvious to callers of `NewImage` or of `Read`, perhaps we could think on this...

luhring (commented):
@wagoodman Thanks for getting this in! No objections to the merge here. Some things to chat about if you're interested

internal/stringset.go
wagoodman: ?

pkg/image/image.go
wagoodman: IMHO the API is intuitive. Read is responsible for populating the original metadata, which means that any additional (overriding) metadata should be applied *after* the read.

I also don't think that `NewImage` should call `Read` implicitly (couldn't tell if that's what you were suggesting?), as this is an intensive operation (in my opinion constructors should rarely do IO, only prep for a future call to do IO)

internal/stringset.go
luhring: Just referring to our discussion/frustration with go's lack of native support for sets and the impetus for rolling your own 😃

pkg/image/image.go
luhring: Agreed, `NewImage` shouldn't call `Read`. And the logic of making these changes after the original metadata is read makes total sense. The documentation could make it clearer what's being passed in and when it gets applied. And it could be that the mutating functions aren't needed from the consumer until the call to `Read` (up for discussion), as opposed to storing the functions on the object in between the calls to `NewImage` and `Read`.

wagoodman: I see that is another way to implement this. Are there specific reasons as to why the current approach is wrong?

PR #54, opened 2020-12-29
Fix link resolution

wagoodman:
Link resolution was only considering leaves of the tree, and not performing link resolution for constituent paths. This PR fixes this behavior by moving content metadata that related to structure (is a dir, link info, file type, etc) into the FileTree and adding helper functions that do link resolution by default (a responsibility previously only done upon fetching content).

Changes:
- `FileTree` has been extracted into its own package
- Added a `FileNode` that leverages the `node.Node` interface as the primary element of the `FileTree` (instead of by `file.Path` with auxiliary information as maps on the `FileTree`... this information now lives on the node). 
- `FileTree.File()` performs link resolution by default.
- Added `LinkResolutionStrategy` (and options) to specify exact link resolution rules.
- Swaps globbing implementation for `doublestar`, where the `FileTree` uses adapters to fulfill the `doublestar.OS` interface.
- Adds `filetree.DepthFirstPathWalker` tailored for iterating paths in `FileTree`s, which may be a superset of the tree nodes (unlike with `tree.DepthFristWalker`, which only traverses tree nodes).

Potential improvements:
- Add more aggressive testing around `FileTree.merge`, this will hint at more validation and merge logic that is needed (e.g. extended functionality + test for when a lower tree has a path with a directory with children that gets overridden with a symlink... the children in the lower layer should be removed).
- Add more integration-level testing around symlink-based edge cases.


pkg/file/path.go
```diff
@@ -72,7 +70,7 @@ func (p Path) ParentPath() (Path, error) {
 	return sanitized, nil
 }
 
-// AllPaths returns all valid constituent paths (e.g. /home/wagoodman/file.txt -> /, /home, /home/wagoodman )
+// AllPaths returns all valid constituent paths (e.g. /home/wagoodman/file.txt -> /, /home, /home/wagoodman, /home/wagoodman/file.txt )
```
luhring, line 73: Re "constituent paths", it'd be good to figure out a more clear way to express which paths get returned. We use "all valid constituent paths" here in the documentation for `AllPaths()` and also for `ConstituentPaths()`, so it's unclear if "constituent" is meant to include the original path itself. Perhaps something like `AllPaths` and `AncestorPaths`?

pkg/file/path.go
```diff
@@ -83,3 +81,20 @@ func (p Path) AllPaths() []Path {
...
+
+// ConstituentPaths returns all valid constituent paths (e.g. /home/wagoodman/file.txt -> /, /home, /home/wagoodman )
+func (p Path) ConstituentPaths() []Path {
+	parents := strings.Split(strings.Trim(string(p), DirSeparator), DirSeparator)
+	fullPaths := make([]Path, len(parents))
+	for idx := range parents {
+		cur := DirSeparator + strings.Join(parents[:idx], DirSeparator)
+		fullPaths[idx] = Path(cur)
+	}
+	return fullPaths
```
luhring, line 93: It'd be good to deduplicate this logic, given the implementation for `AllPaths`. One suggestion would be to have `AllPaths` consume this method and append the original path.

pkg/file/path_queue.go
```diff
@@ -0,0 +1,43 @@
+package file
+
+type PathQueue struct {
```
luhring, line 3: I couldn't find where this is getting used, is this something we need?

pkg/filetree/filetree.go
```diff
@@ -0,0 +1,749 @@
...
+
+// Reader returns a tree.Reader useful for Tree traversal.
+func (t *FileTree) Reader() tree.Reader {
+	return t.tree
+}
+
+// PathDiff shows the path differences between two trees (useful for testing)
+func (t *FileTree) PathDiff(other *FileTree) (extra, missing []file.Path) {
+	extra = make([]file.Path, 0)
+	missing = make([]file.Path, 0)
```
luhring, line 616: These two lines ( for `extra` and `missing`) might be unnecessary, since no capacity is allocated, and these variables are already declared via the return parameters.

pkg/filetree/glob.go
```diff
@@ -0,0 +1,192 @@
...
+	// by making the file point to the same inode as another --making the indirection transparent to applications.
+	if a.Node.FileType == file.TypeSymlink || a.Node.FileType == file.TypeHardLink {
+		mode |= os.ModeSymlink
+	}
+	return mode
+}
+
+// ModTime returns a dummy value. Traditionally would be the modification time for the given file.
+func (a *fileinfoAdapter) ModTime() time.Time {
+	return time.Now()
```
luhring, line 181: (Nit) Interesting... if the return value really doesn't matter for our code, and we know we won't have a "real" value to offer to the consumer, I might suggest making this implementation deterministic to avoid the potential for bugs in future consumption of this method.

luhring (commented):
This is a lot, but so far I'm liking the new direction. There's probably even more to evolve as we come across more use cases and reflect on how this approach is working.

I left just a few comments, but also happy to take over and drive if you get pulled away unexpectedly! 😃

go.sum
```diff
@@ -131,6 +131,12 @@ github.com/bgentry/speakeasy v0.1.0/go.mod h1:+zsyZBPWlz7T6j88CTgSN5bM796AkVf0kB
 github.com/bketelsen/crypt v0.0.3-0.20200106085610-5cbc8cc4026c/go.mod h1:MKsuJmJgSg28kpZDP6UIiPt0e0Oz0kqKNGyRaWEPv84=
 github.com/blakesmith/ar v0.0.0-20190502131153-809d4375e1fb/go.mod h1:PkYb9DJNAwrSvRx5DYA+gUcOIgTGVMNkfSCbZM8cWpI=
 github.com/blang/semver v3.5.0+incompatible/go.mod h1:kRBLl5iJ+tD4TcOOxsy/0fnwebNt5EWlYSAyrTnjyyk=
+github.com/bmatcuk/doublestar v1.3.1 h1:rT8rxDPsavp9G+4ZULzqhhUSaI/OPsTZNG88Z3i0xvY=
+github.com/bmatcuk/doublestar v1.3.1/go.mod h1:wiQtGV+rzVYxB7WIlirSN++5HPtPlXEo9MEoZQC/PmE=
+github.com/bmatcuk/doublestar v1.3.4 h1:gPypJ5xD31uhX6Tf54sDPUOBXTqKH4c9aPY66CyQrS0=
+github.com/bmatcuk/doublestar v1.3.4/go.mod h1:wiQtGV+rzVYxB7WIlirSN++5HPtPlXEo9MEoZQC/PmE=
+github.com/bmatcuk/doublestar/v2 v2.0.4 h1:6I6oUiT/sU27eE2OFcWqBhL1SwjyvQuOssxT4a1yidI=
+github.com/bmatcuk/doublestar/v2 v2.0.4/go.mod h1:QMmcs3H2AUQICWhfzLXz+IYln8lRQmTZRptLie8RgRw=
```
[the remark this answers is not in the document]
wagoodman: I was accidentally at one point, but switched to v2, I'll double check and remove these lines

pkg/filetree/glob.go
wagoodman: in actuality any usage of this function should be considered unsupported. Should probably even replace these with panics?

pkg/file/path_queue.go
wagoodman: It would have been used in the BFS, I implemented it out of habit. I'll remove.

pkg/file/path.go
wagoodman: I meant to do exactly that, I'll adjust

pkg/file/path.go
wagoodman: I think the description needs to be updated (it is currently wrong for both functions), I can adjust.

go.sum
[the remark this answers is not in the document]
wagoodman: confirmed, we are only using v2

pkg/filetree/glob.go
wagoodman: I've adjusted all unused methods that are there to fulfill the interface with panic statements.

luhring: I think that's a good solution 👍

pkg/filetree/depth_first_path_walker_test.go
```diff
@@ -0,0 +1,287 @@
...
+	for _, p := range tailPaths {
+		for _, c := range file.Path(p).ConstituentPaths() {
+			if c == "/" {
+				continue
+			}
+			delete(possiblePaths, string(c))
+		}
+		delete(possiblePaths, p)
+	}
+	// </delete>
```
alfredodeza, line 138: left over comment? or actionable?

alfredodeza: ooooh never mind, I think this means that this whole block is for deleting. Perhaps makes sense to extract it out? (nit for sure)

pkg/filetree/depth_first_path_walker_test.go
```diff
@@ -0,0 +1,287 @@
...
+
+	actualPaths := make(map[string]*file.Reference, 0)
+	visitor := func(path file.Path, node filenode.FileNode) error {
```
alfredodeza, line 141: This is probably me, but how would this ever return an error?

pkg/filetree/depth_first_path_walker_test.go
```diff
@@ -0,0 +1,287 @@
...
+	if actualMaxDepth == -1 || actualMaxDepth > maxDirDepth {
+		t.Fatalf("never traversed or went above allowable threshold")
+	}
+
+}
+
+func assertExpectedTraversal(t *testing.T, expected, actual map[string]*file.Reference) {
+	t.Helper()
+	if len(actual) != len(expected) {
+		t.Errorf("Did not traverse all nodes (expected %d, got %d)", len(expected), len(actual))
```
alfredodeza, line 281: tiny nit: the comparison is done with `actual` to `expected` but the messaging is `expected` to `actual`. I don't have a strong preference but consistency would be nice, especially when there are failures

pkg/filetree/filetree.go
```diff
@@ -0,0 +1,754 @@
...
+		return nil, err
+	}
+
+	if n == nil {
+		return nil, nil
+	}
+
+	if n.FileType != file.TypeDir {
+		return nil, nil
+	}
```
alfredodeza, line 83: wouldn't it be easier for the consumer of this method to get back an empty slice of Path (`[].file.Path`) instead of a `nil`? The method definition here is about listing paths, and from what I can see how this is consumed, the consumers aren't checking if they get a `nil` back

pkg/filetree/filetree.go
```diff
@@ -0,0 +1,754 @@
...
+func (t *FileTree) resolveAncestorLinks(path file.Path) (*filenode.FileNode, error) {
+	var pathParts = strings.Split(string(path.Normalize()), file.DirSeparator)
+	var currentPathStr string
+	var currentPath file.Path
+	var currentNode *filenode.FileNode
+	var err error
+
+	// iterate through all parts of the path, replacing path elements with link resolutions where possible.
+	for idx, part := range pathParts {
+		if (part == "" || part == file.DirSeparator) && idx == 0 {
```
alfredodeza, line 192: It seems that these conditional checks would be equal to checking if the `part == ""`. The rest of the conditions read somewhat redundant. Are they really necessary? Or in what situations the root path `"/"` would produce something other than an empty string?

pkg/filetree/depth_first_path_walker_test.go
wagoodman: this could have been more clear... the above code deletes paths that are not expected from the expected test fixture.... I'll update the comment

pkg/filetree/depth_first_path_walker_test.go
wagoodman: indeed, the walker.Walk() will return the error if the visitor returns an error https://github.com/anchore/stereoscope/blob/81d38ffbdd655032a66d430d83a923ff07fb9db1/pkg/filetree/depth_first_path_walker.go#L92-L95

pkg/filetree/filetree.go
wagoodman: this is one of the nice features of go, where a `nil` slice is the same as an empty slice (`[]file.Path{}` or `make([]file.Path, 0)`)

alfredodeza (approved):

PR #57, opened 2021-02-01
Fix zero layers panic

### LINKED ISSUE #56: Panic on images with zero layers
Stereoscope panics when returning results for an image that has no layers (e.g. scratch). This can be seen by running the following from stereoscope's root directory:

```
skopeo --override-os linux copy docker://anchore/test_images:scratch_v1 oci:./test-scratch
go run ./examples/basic.go oci-dir:./test-scratch
```

The cause of this panic is here: https://github.com/anchore/stereoscope/blob/main/pkg/image/image.go#L200

This code assumes there is at least one layer in the image and thus doesn't check to make sure `len(i.Layers)-1` produces a valid index.

**NOTE:** If you try to use Docker Engine to save the image you'll get a _totally different_ error, as seen here:

```
go run ./examples/basic.go anchore/test_images:scratch_v1
panic: unable to save image tar: Error response from daemon: empty export - not implemented
```

luhring:
Fixes #56 by checking for the case of zero layers before returning a squashed tree

### DIFF
--- a/pkg/image/image.go
+++ b/pkg/image/image.go
@@ -197,7 +197,14 @@ func (i *Image) squash(prog *progress.Manual) error {
 
 // SquashedTree returns the pre-computed image squash file tree.
 func (i *Image) SquashedTree() *filetree.FileTree {
-	return i.Layers[len(i.Layers)-1].SquashedTree
+	layerCount := len(i.Layers)
+
+	if layerCount == 0 {
+		return filetree.NewFileTree()
+	}
+
+	topLayer := i.Layers[layerCount-1]
+	return topLayer.SquashedTree
 }
 
 // FileContentsFromSquash fetches file contents for a single path, relative to the image squash tree.
--- a/pkg/image/image_test.go
+++ b/pkg/image/image_test.go
@@ -3,11 +3,12 @@ package image
 import (
 	"crypto/sha256"
 	"fmt"
-	"github.com/go-test/deep"
-	"github.com/google/go-containerregistry/pkg/name"
 	"io/ioutil"
 	"os"
 	"testing"
+
+	"github.com/go-test/deep"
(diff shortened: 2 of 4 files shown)


pkg/image/image_test.go
```diff
@@ -96,3 +97,24 @@ func TestImageAdditionalMetadata(t *testing.T) {
...
+			if r := recover(); r != nil {
+				t.Errorf("panicked (and recovered) while computing squashed tree for image with zero layers: %v", r)
+			}
+		}()
+
+		// Asserting that this call doesn't panic (regression: https://github.com/anchore/stereoscope/issues/56)
+		result := i.SquashedTree()
+
+		if result == nil {
+			t.Error("expected an initialized, empty FileTree, but got a nil FileTree")
```
alfredodeza, line 117: (probably a nit), but is there such a thing like a `nil FileTree` ? I thought that this would be a (plain) `nil`
```suggestion
			t.Error("expected an initialized, empty FileTree, but got nil")
```

pkg/image/image.go
```diff
@@ -197,7 +197,14 @@ func (i *Image) squash(prog *progress.Manual) error {
 
 // SquashedTree returns the pre-computed image squash file tree.
 func (i *Image) SquashedTree() *filetree.FileTree {
-	return i.Layers[len(i.Layers)-1].SquashedTree
+	layerCount := len(i.Layers)
```
alfredodeza, line 200: thanks for splitting these up into meaningful variables

alfredodeza (approved):

pkg/image/image_test.go
luhring: @alfredodeza Good question — Actually, all instances of `nil` in Go _are associated_ with a specific type. So it actually is a "nil FileTree". (This is different from `null` in other languages)

PR #58, opened 2021-03-15
Add TarIndex for more efficient content fetching

wagoodman:
This PR replaces the existing batch content fetch approach with a one-pass tar index builder approach. Subsequent content requests are given a `DeferredPartialReadCloser` which:
- upon first read will open the (uncompressed) tar file.
- start reading at the payload start position and prevent reading further than what the tar header indicates (by size)
- upon reading EOF will automatically close the file (to reduce unnecessary file descriptor usage).

Upon reading the image, each tar entry is used to create a `file.Opener` which is persisted onto the `FileCatalogEntry` for that file, acting as a factory of  `DeferredPartialReadCloser` for individual file contents.

This PR also adds the first benchmark test (more to come later). On each opened PR comparative benchmark testing is run against the last results from the main branch and added as a sticky comment on the PR. If the main branch results are not available (as will be the case with this PR) then only the raw stats are shown.

Note: the consumers will create a file descriptor for each file actively read, but if not closed or the contents full read, will lead to leaking file descriptors. There is no way to limit how many descriptors are handed out at this time, which means that consumers will need to be diligent here.

Additionally this PR moves the intergration tests to the `test/` directory, as other toolbox repos do.


.github/workflows/basic-checks.yaml
```diff
@@ -8,6 +8,7 @@ jobs:
   Static-Analysis:
     strategy:
       matrix:
+        # note: bumping the go value should be done for all workflows. Make certain to check conditionals as well.
```
luhring, line 11: What does "go value" refer to here? The version?

.github/workflows/basic-checks.yaml
```diff
@@ -42,6 +43,7 @@ jobs:
     strategy:
       matrix:
         # test the lower bounds of support, and the latest available
+        # note: bumping the go value should be done for all workflows. Make certain to check conditionals as well.
```
luhring, line 46: (ditto above comment)

pkg/file/deferred_partial_read_closer.go
```diff
@@ -0,0 +1,73 @@
...
+var _ io.ReadCloser = (*deferredPartialReadCloser)(nil)
+
+// deferredPartialReadCloser is a "lazy" read closer, allocating a file descriptor for the given path only upon the first Read() call.
+// Additionally only part of the file is allowed to be read, starting at a given position.
+type deferredPartialReadCloser struct {
+	// path is the path to be opened
+	path string
+	// file is the io.ReadCloser source for the path
+	file   *os.File
+	reader io.Reader
```
luhring, line 18: I think it'd be helpful to explain the function of `reader` as well

pkg/file/deferred_partial_read_closer.go
```diff
@@ -0,0 +1,73 @@
...
+	"errors"
+	"io"
+	"os"
+)
+
+var _ io.ReadCloser = (*deferredPartialReadCloser)(nil)
+
+// deferredPartialReadCloser is a "lazy" read closer, allocating a file descriptor for the given path only upon the first Read() call.
+// Additionally only part of the file is allowed to be read, starting at a given position.
+type deferredPartialReadCloser struct {
```
luhring, line 13: nit / for consideration: The name `deferredPartialReadCloser` is not immediately obvious, but I see the motivation, especially for the descriptors "deferred" and "partial". It might be good to consider other descriptors in an attempt to increases the obviousness. For example, once I saw "lazy" in your comment, I immediately grasped that concept. Maybe "lazy" in place of "deferred"? (Also "defer" has a particular connotation in Go.) For "partial", my impression based on the implementation is that it's not necessarily partial — the consumer can specify any starting point and size, including values such that it's actually the whole file. I'm thinking "bounded" or something might be closer? What do you think?

luhring: Once I understood the concept, though, I loved it. This "object" is a very useful construct. ❤️

pkg/file/deferred_partial_read_closer.go
```diff
@@ -0,0 +1,73 @@
...
+
+		d.file = file
+		d.reader = io.LimitReader(d.file, d.size)
+	}
+	n, err := d.reader.Read(b)
+	if err != nil && errors.Is(err, io.EOF) {
+		// we've reached the end of the file, force a release of the file descriptor. If the file has already been
+		// closed, ignore the error.
+		if closeErr := d.file.Close(); !errors.Is(closeErr, os.ErrClosed) {
+			err = closeErr
```
luhring, line 53: Is it possible to return early here?

pkg/file/metadata.go
```diff
@@ -1,13 +1,18 @@
...
+	"path"
+)
 
 // Metadata represents all file metadata of interest (used today for in-tar file resolution).
 type Metadata struct {
 	// Path is the absolute path representation to the file
 	Path string
 	// TarHeaderName is the exact entry name as found within a tar header
 	TarHeaderName string
+	TarSequence   int64
```
luhring, line 15: It'd be helpful to document the `TarSequence` field

pkg/file/tar_index.go
```diff
@@ -0,0 +1,70 @@
...
+func (t *TarIndex) indexEntries(file *os.File, onIndex ...TarIndexVisitor) error {
+	visitor := func(entry TarFileEntry) error {
+		payloadLocation, err := file.Seek(0, io.SeekCurrent)
+		if err != nil {
+			return fmt.Errorf("unable to read current position in tar: %v", err)
+		}
+
+		// keep track of the header position for this entry; the current file position is where the entry
+		// body payload starts (after the header has been read).
+		index := TarIndexEntry{
```
luhring, line 39: It'd be more clear to call this variable something like `entry`, to avoid confusion around what `index` means in this context.

pkg/file/tar_index.go
```diff
@@ -0,0 +1,70 @@
...
+func NewTarIndex(tarFilePath string, onIndex ...TarIndexVisitor) (*TarIndex, error) {
+	t := &TarIndex{
+		indexByName: make(map[string][]TarIndexEntry),
+	}
+	fh, err := os.Open(tarFilePath)
+	if err != nil {
+		return nil, err
+	}
+	defer fh.Close()
+	return t, t.indexEntries(fh, onIndex...)
```
luhring, line 26: I'm not used to seeing an object's method be used in its own construction in OOP, but I like that the indexing functionality was broken out into its own function. Is there something that necessitated this approach?

pkg/file/tar_index.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// indexEntries records all tar header locations indexed by header names.
+func (t *TarIndex) indexEntries(file *os.File, onIndex ...TarIndexVisitor) error {
```
luhring, line 30: It could be helpful to somehow indicate in the name of the `file` parameter that this is the TAR file, just to be clear that it's not a particular file contained in the TAR.

pkg/file/tarutil.go
```diff
@@ -40,19 +43,28 @@ func (e *ErrFileNotFound) Error() string {
 // TarIterator is a function that reads across a tar and invokes a visitor function for each entry discovered. The iterator
 // stops when there are no more entries to read, if there is an error in the underlying reader or visitor function,
 // or if the visitor function returns a ErrTarStopIteration sentinel error.
-func TarIterator(reader io.Reader, visitor TarVisitor) error {
+func TarIterator(reader io.Reader, visitor TarFileVisitor) error {
```
luhring, line 46: I know this isn't a change in this PR, but here's a comment nonetheless 😄  — it'd be nice to improve the name `TarIterator`, specifically in two ways: 1) It's usually easier to read functions in an implementation when there's a verb-like phrase instead of a noun, and this would be an opportunity to explain what the function is doing; and 2) I'm used to "iterators" being objects or functions that produce values iteratively, as opposed to the entity that's asking for new values, so the name sets an expectation for me that's not met in the function's implementation.

pkg/image/content_helpers.go
```diff
@@ -26,26 +25,3 @@ func fetchFileContentsByPath(ft *filetree.FileTree, fileCatalog *FileCatalog, pa
 	}
 	return reader, nil
 }
-
-// fetchMultipleFileContentsByPath is a common helper function for resolving the file contents for all paths from the
-// file catalog relative to the given tree. If any one path does not exist in the given tree then an error is returned.
-func fetchMultipleFileContentsByPath(ft *filetree.FileTree, fileCatalog *FileCatalog, paths ...file.Path) (map[file.Reference]io.ReadCloser, error) {
```
luhring, line 32: Nice cleanup!!!

luhring (commented):
This is a fantastic direction for Stereoscope! 🎉  Made a few notes to talk through...

.github/workflows/basic-checks.yaml
wagoodman: yeah, I'll update to make more clear

pkg/file/deferred_partial_read_closer.go
wagoodman: 👍  went with "LazyBoundedReadCloser" and also renamed it's sibling "LazyReadCloser"

pkg/file/deferred_partial_read_closer.go
wagoodman: could go either way here, so I'll make it an early return

pkg/file/tar_index.go
wagoodman: when this was a `IndexedTarReader` implementation it made a lot of sense, but with the smaller implementation as a `TarIndex` it doesn't make much sense --I'll merge both of these functions together.

pkg/file/lazy_bounded_read_closer.go
```diff
@@ -6,31 +6,32 @@ import (
...
-// deferredPartialReadCloser is a "lazy" read closer, allocating a file descriptor for the given path only upon the first Read() call.
+// lazyBoundedReadCloser is a "lazy" read closer, allocating a file descriptor for the given path only upon the first Read() call.
 // Additionally only part of the file is allowed to be read, starting at a given position.
-type deferredPartialReadCloser struct {
+type lazyBoundedReadCloser struct {
 	// path is the path to be opened
 	path string
-	// file is the io.ReadCloser source for the path
-	file   *os.File
+	// file is active file handle for the given path
```
luhring, line 16: nit: "file is **_the_** active file handle" ?

luhring (approved):
Looks great!

PR #60, opened 2021-03-23
Add optional file.Type filter to FileTree.AllFiles call

wagoodman:
This enables callers to enumerate files of a particular type (defaults to regular files).


pkg/filetree/filetree_test.go
```diff
@@ -786,3 +786,94 @@ func TestFileTree_File_CycleDetection(t *testing.T) {
...
+			t.Fatalf("failed to add path ('%s'): %+v", p, err)
+		}
+	}
+
+	var err error
+
+	// dir
+	_, err = tr.AddDir("/home")
+	if err != nil {
+		t.Fatalf("could not setup link: %+v", err)
```
luhring, line 811: What does "could not setup link" mean here?

pkg/filetree/filetree.go
```diff
@@ -43,11 +43,20 @@ func (t *FileTree) Copy() (*FileTree, error) {
 }
 
 // AllFiles returns all files and directories within the FileTree.
-func (t *FileTree) AllFiles() []file.Reference {
+func (t *FileTree) AllFiles(types ...file.Type) []file.Reference {
```
luhring, line 46: It would be nice to document type filtering behavior and the "default" behavior (no supplied types) in the doc comment.

pkg/filetree/filetree_test.go
```diff
@@ -786,3 +786,94 @@ func TestFileTree_File_CycleDetection(t *testing.T) {
...
+			types: []file.Type{file.TypeDir},
+			// note: only explicitly added directories exist in the catalog
+			expected: internal.NewStringSet("/home"),
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			actual := tr.AllFiles(test.types...)
+			if len(actual) != len(test.expected) {
```
luhring, line 867: It might be nice to check out a comparison assertion from Testify to make this check more concise and expressive

pkg/filetree/filetree_test.go
wagoodman: should say dir, I'll adjust

pkg/filetree/filetree_test.go
wagoodman: adjusting to `assert.Contains`

luhring (approved):
Looks good!

PR #62, opened 2021-04-03
Add field logger to the logging interface

wagoodman:
There are several instances of key-value parameters that are captured into the log message. Since syft and grype support structured logging it would be ideal to capture these values outside of the log message. Additionally, when kept separate these fields can be individually formatted when using standard logging.

Additionally found and removed an instance of apex/log.

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -1,7 +1,19 @@
 package logger
 
 type Logger interface {
+	MessageLogger
+	FieldLogger
+}
+
+type FieldLogger interface {
+	WithFields(fields ...interface{}) MessageLogger
+}
+
+type MessageLogger interface {
 	Errorf(format string, args ...interface{})
+	Error(args ...interface{})
+	Warnf(format string, args ...interface{})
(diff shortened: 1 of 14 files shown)


wagoodman:
Interesting... gonna wait this one out since this seems like a dockerhub problem

Screen Shot 2021-04-03 at 9 14 51 AM

pkg/logger/logger.go
```diff
@@ -1,7 +1,19 @@
 package logger
 
 type Logger interface {
+	MessageLogger
+	FieldLogger
+}
+
+type FieldLogger interface {
+	WithFields(fields ...interface{}) MessageLogger
```
wagoodman, line 9: the problem with this approach is that the `MessageLogger` is part of the return type, meaning that if a consumer wanted to bring their own logger, that logger would need to have this interface in the signature, which wont scale.

wagoodman:
This wont work in a pluggable fashion, not without centralizing the logging facilities into it's own lib, something we don't really need to do yet (but not against it either).

PR #63, opened 2021-04-05
Ensure that credentials are not HTML encoded

wagoodman:
Today when setting docker pull options we are responsible for base64 encoding the `RegistryAuth` JSON, which is an object containing username and password. Before base64 encoding we are not making certain that the JSON marshaling will not escape specific characters typical to HTML documents. This PR swaps out the JSON unmarshal with a JSON encoder with HTML escaping turned off.

Related to https://github.com/anchore/grype/issues/254

### DIFF
--- a/pkg/image/docker/daemon_provider.go
+++ b/pkg/image/docker/daemon_provider.go
@@ -1,6 +1,7 @@
 package docker
 
 import (
+	"bytes"
 	"context"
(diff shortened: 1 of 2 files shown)


luhring (approved):
Nice! 🤘

alfredodeza (approved):

PR #64, opened 2021-04-06
Add registry image source

wagoodman:
Adds the `oci-registry` / `registry` image source, enabling the ability to pull images directly from a registry (not through the docker daemon). This brings with it the requirement to provide credentials in the form of username/password or token.

What this PR does _**NOT**_ bring in yet it credential helpers for common registries (see #65 ):
 - ECR (via https://github.com/awslabs/amazon-ecr-credential-helper)
 - GCR (via https://github.com/GoogleCloudPlatform/docker-credential-gcr)
 - ACR (via https://github.com/Azure/acr-docker-credential-helper)

Addresses https://github.com/anchore/grype/issues/264

### DIFF
--- a/pkg/image/oci/registry_provider.go
+++ b/pkg/image/oci/registry_provider.go
@@ -60,33 +60,17 @@ func prepareRemoteOptions(ref name.Reference, registryOptions *image.RegistryOpt
 		}
 		opts = append(opts, remote.WithTransport(t))
 	}
-	registry := ref.Context().RegistryStr()
-	discoveredConfiguredAuth := false
-	for idx, auth := range registryOptions.Credentials {
-		if auth.Authority == registry {
-			if auth.Username != "" && auth.Password != "" {
-				log.Debugf("using registry credentials for %q (config idx=%d)", auth.Authority, idx)
-				opts = append(opts, remote.WithAuth(&authn.Basic{
-					Username: auth.Username,
-					Password: auth.Password,
-				}))
-				discoveredConfiguredAuth = true
-				break
-			} else if auth.Token != "" {
-				log.Debugf("using registry token for %q (config idx=%d)", auth.Authority, idx)
-				opts = append(opts, remote.WithAuth(&authn.Bearer{
-					Token: auth.Token,
-				}))
-				discoveredConfiguredAuth = true
-				break
-			}
-		}
-	}
 
-	if !discoveredConfiguredAuth {
+	// note: the authn.Authenticator and authn.Keychain options are mutually exclusive, only one may be provided.
+	// If no explicit authenticator can be found, then fallback to the keychain.
+	authenticator := registryOptions.Authenticator(ref.Context().RegistryStr())
+	if authenticator != nil {
+		opts = append(opts, remote.WithAuth(authenticator))
+	} else {
 		// use the Keychain specified from a docker config file.
 		log.Debugf("using registry credentials from default keychain")
 		opts = append(opts, remote.WithAuthFromKeychain(authn.DefaultKeychain))
 	}
+
 	return opts
 }
--- a/pkg/image/oci/registry_provider.go
+++ b/pkg/image/oci/registry_provider.go
@@ -1,0 +1,92 @@
+package oci
+
+import (
+	"crypto/tls"
+	"fmt"
+	"net/http"
+
+	"github.com/anchore/stereoscope/internal/log"
+	"github.com/anchore/stereoscope/pkg/file"
+	"github.com/anchore/stereoscope/pkg/image"
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
+
+// RegistryImageProvider is a image.Provider capable of fetching and representing a container image fetched from a remote registry (described by the OCI distribution spec).
+type RegistryImageProvider struct {
+	imageStr        string
+	tmpDirGen       *file.TempDirGenerator
+	registryOptions *image.RegistryOptions
+}
+
+// NewProviderFromRegistry creates a new provider instance for a specific image that will later be cached to the given directory.
+func NewProviderFromRegistry(imgStr string, tmpDirGen *file.TempDirGenerator, registryOptions *image.RegistryOptions) *RegistryImageProvider {
+	return &RegistryImageProvider{
+		imageStr:        imgStr,
+		tmpDirGen:       tmpDirGen,
+		registryOptions: registryOptions,
+	}
+}
+
+// Provide an image object that represents the cached docker image tar fetched a registry.
+func (p *RegistryImageProvider) Provide() (*image.Image, error) {
+	log.Debugf("pulling image info directly from registry image=%q", p.imageStr)
+
+	imageTempDir, err := p.tmpDirGen.NewTempDir()
+	if err != nil {
+		return nil, err
+	}
+
+	ref, err := name.ParseReference(p.imageStr)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse registry reference=%q: %+v", p.imageStr, err)
+	}
+
+	img, err := remote.Image(ref, prepareRemoteOptions(ref, p.registryOptions)...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create image from registry: %+v", err)
+	}
+
+	return image.NewImage(img, imageTempDir), nil
+}
+
+func prepareRemoteOptions(ref name.Reference, registryOptions *image.RegistryOptions) []remote.Option {
+	var opts []remote.Option
+	if registryOptions.InsecureSkipTLSVerify {
+		t := &http.Transport{
+			// nolint: gosec
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+		opts = append(opts, remote.WithTransport(t))
+	}
+	registry := ref.Context().RegistryStr()
+	discoveredConfiguredAuth := false
+	for idx, auth := range registryOptions.Auth {
+		if auth.Authority == registry {
+			if auth.Username != "" && auth.Password != "" {
+				log.Debugf("using registry credentials for %q (config idx=%d)", auth.Authority, idx)
+				opts = append(opts, remote.WithAuth(&authn.Basic{
+					Username: auth.Username,
+					Password: auth.Password,
+				}))
+				discoveredConfiguredAuth = true
+				break
+			} else if auth.Token != "" {
+				log.Debugf("using registry token for %q (config idx=%d)", auth.Authority, idx)
+				opts = append(opts, remote.WithAuth(&authn.Bearer{
+					Token: auth.Token,
+				}))
+				discoveredConfiguredAuth = true
+				break
+			}
+		}
+	}
+
+	if !discoveredConfiguredAuth {
+		// use the Keychain specified from a docker config file.
+		log.Debugf("using registry credentials from default keychain")
+		opts = append(opts, remote.WithAuthFromKeychain(authn.DefaultKeychain))
+	}
+	return opts
+}
--- a/pkg/image/oci/registry_provider.go
+++ b/pkg/image/oci/registry_provider.go
@@ -68,7 +68,7 @@ func prepareRemoteOptions(ref name.Reference, registryOptions *image.RegistryOpt
 		opts = append(opts, remote.WithAuth(authenticator))
 	} else {
 		// use the Keychain specified from a docker config file.
-		log.Debugf("using registry credentials from default keychain")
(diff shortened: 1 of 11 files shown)


pkg/image/oci/registry_provider.go
```diff
@@ -0,0 +1,92 @@
...
+
+// RegistryImageProvider is a image.Provider capable of fetching and representing a container image fetched from a remote registry (described by the OCI distribution spec).
+type RegistryImageProvider struct {
+	imageStr        string
+	tmpDirGen       *file.TempDirGenerator
+	registryOptions *image.RegistryOptions
+}
+
+// NewProviderFromRegistry creates a new provider instance for a specific image that will later be cached to the given directory.
+func NewProviderFromRegistry(imgStr string, tmpDirGen *file.TempDirGenerator, registryOptions *image.RegistryOptions) *RegistryImageProvider {
```
alfredodeza, line 24: Why is this called `NewProviderFromRegistry`? I thought the convention was to name it like the struct it creates?

wagoodman: good call, I'll adjust 👍

pkg/image/registry_options.go
```diff
@@ -0,0 +1,16 @@
+package image
+
+// RegistryOptions for the OCI registry provider.
+type RegistryOptions struct {
+	InsecureSkipTLSVerify bool
+	Auth                  []RegistryCredentials
```
alfredodeza, line 6: Since this key is singular, it complicates following the code when it is used in a loop. Perhaps `AuthenticationMethods` or if the struct was called `Authentication` this could be `Methods`? Again, no strong suggestions, but as they are presented here, makes it hard to follow them when they are used

pkg/image/oci/registry_provider.go
```diff
@@ -0,0 +1,92 @@
...
+		t := &http.Transport{
+			// nolint: gosec
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+		opts = append(opts, remote.WithTransport(t))
+	}
+	registry := ref.Context().RegistryStr()
+	discoveredConfiguredAuth := false
+	for idx, auth := range registryOptions.Auth {
+		if auth.Authority == registry {
```
alfredodeza, line 66: why does it need to go through all the authentication methods to return all valid ones instead of the first one?

pkg/image/oci/registry_provider.go
```diff
@@ -0,0 +1,92 @@
...
+			if auth.Username != "" && auth.Password != "" {
+				log.Debugf("using registry credentials for %q (config idx=%d)", auth.Authority, idx)
```
alfredodeza, line 68: this block here is not "using registry credentials" it found them and appended them to return them to something else that will use them

pkg/image/registry_options.go
```diff
@@ -0,0 +1,16 @@
+package image
+
+// RegistryOptions for the OCI registry provider.
```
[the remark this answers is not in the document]
wagoodman: Currently this is supporting a little more than auth --the `InsecureSkipTLSVerify ` option is for the http transport beyond just auth (includes any http interaction). I'm up for any name change here, but wanted to point out the intended scope for this object.

pkg/image/registry_options.go
wagoodman: agreed, I think we can do better than `Auth`. How about changing the field name from `Auth` to `Credentials`? That would solve both the plurality problem and more closely align with the struct name (`RegistryCredentials`) without being redundant on the word "Registry".

pkg/image/oci/registry_provider.go
wagoodman: This code is currently `break`ing when it finds a valid authentication configuration (matching authority and basic vs bearer token selection). But I should at the very least come back through here and add more comments, mentioning some of the necessary behavior (e.g. these authentication-based options are mutually exclusive options).

This function as a whole is meant to prepare all `remote.Options` which goes beyond only authentication options (there's the transport option at the top of this function).

pkg/image/oci/registry_provider.go
wagoodman: I think https://github.com/anchore/stereoscope/pull/64#discussion_r608828299 may answer this, but let me know if it doesn't 👍

pkg/image/registry_options.go
wagoodman: done deal in the latest commit

pkg/image/oci/registry_provider.go
alfredodeza: I see that this loop breaks, but instead of looping + appending + breaking, why can't it return instead of a break condition and omit appending? 
When you say "Options goes beyond only authentication" it becomes problematic, because this is treating authentication in the same block as a TLS setting. It seems to me that this would be easier to grasp if connection settings are separate from authentication, which would avoid these constructs for handling them all in one loop.

wagoodman: re: "options go beyond auth", these options are from the gcr lib and we don't have a lot of control when it comes to this form factor... we still have to prepare a list of `remote.Option`.

What I can do to make this easier to read is break out the auth options into their own function, which should simplify that logic some by returning early as you are suggesting. Just a note, `prepareRemoteOptions` would still be prepping the TLS config / http transport but would also be additionally invoking this new function dedicated to auth and appending any new options to the existing options list.

pkg/image/registry_options.go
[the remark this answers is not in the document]
alfredodeza: If a user can set TLS for any registry interaction, that may be tricky if supporting one registry that needs TLS and another one that doesn't. Are you implying that the TLS option shouldn't change and be global for any/all registries?

pkg/image/oci/registry_provider.go
wagoodman: on the latest commit 👍

pkg/image/registry_options_test.go
```diff
@@ -0,0 +1,206 @@
...
+func TestRegistryOptions_Authenticator_Bearer(t *testing.T) {
+	tests := []struct {
+		name     string
+		registry string
+		input    *RegistryOptions
+		expected *authn.Bearer
+	}{
+		{
+			name:     "require bearer token success",
+			registry: "localhost:5000",
```
alfredodeza, line 165: maybe overkill but you could add a case for no matching registry just like the basic auth

pkg/image/source.go
```diff
@@ -105,9 +113,21 @@ func detectSource(fs afero.Fs, userInput string) (Source, string, error) {
...
-			return DockerDaemonSource, userInput, nil
+		// ignore any source hint since the source is ultimately unknown, see if this could be a docker image
+		if isRegistryReference(userInput) {
+			// verify that the docker daemon is accessible before assuming we can suggest to use it
+			dockerClient, err := docker.GetClient()
+			if err == nil {
+				ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+				defer cancel()
+				pong, err := dockerClient.Ping(ctx)
+				if err == nil && pong.APIVersion != "" {
```
alfredodeza, line 124: is it possible to have a registry that doesn't return an APIVersion?

pkg/image/source.go
```diff
@@ -105,9 +113,21 @@ func detectSource(fs afero.Fs, userInput string) (Source, string, error) {
...
 	case UnknownSource:
-		if isDockerReference(userInput) {
-			// ignore any source hint since the source is ultimately unknown, see if this could be a docker image
-			return DockerDaemonSource, userInput, nil
+		// ignore any source hint since the source is ultimately unknown, see if this could be a docker image
+		if isRegistryReference(userInput) {
+			// verify that the docker daemon is accessible before assuming we can suggest to use it
+			dockerClient, err := docker.GetClient()
+			if err == nil {
+				ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
```
alfredodeza, line 121: I missed this, but why is the timeout 1 second? I can see how this could break and cause issues with no way to workaround it. In corporate environments with VPNs required to access remote servers I would expect this to take more than a second

pkg/image/registry_options.go
```diff
@@ -0,0 +1,44 @@
...
+	Username  string
+	Password  string
+	Token     string
+}
+
+// Authenticator returns an object capable of authenticating against the given registry. If no credentials match the
+// given registry, or there is partial information configured, then nil is returned.
+func (r *RegistryOptions) Authenticator(registry string) authn.Authenticator {
+	for idx, auth := range r.Credentials {
+		if auth.Authority == registry {
```
alfredodeza, line 27: nit: _I_ tend to avoid nesting this much (for -> if -> if || else if) because it makes it trickier to read. One workaround is returning (continuing here I guess) when this condition _is not met_, which would allow the rest of the code to execute. That would outdent the rest, making it easier to read. Again, just a nit.

pkg/image/oci/registry_provider.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create image from registry: %+v", err)
+	}
+
+	return image.NewImage(img, imageTempDir), nil
+}
+
+func prepareRemoteOptions(ref name.Reference, registryOptions *image.RegistryOptions) []remote.Option {
+	var opts []remote.Option
+	if registryOptions.InsecureSkipTLSVerify {
```
alfredodeza, line 56: nit: if the configuration of `tls.Config` is `InsecureSkipVerify` it would be easier to reason about this if it is named the same

pkg/image/oci/registry_provider.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	// note: the authn.Authenticator and authn.Keychain options are mutually exclusive, only one may be provided.
+	// If no explicit authenticator can be found, then fallback to the keychain.
+	authenticator := registryOptions.Authenticator(ref.Context().RegistryStr())
+	if authenticator != nil {
+		opts = append(opts, remote.WithAuth(authenticator))
+	} else {
+		// use the Keychain specified from a docker config file.
+		log.Debugf("using registry credentials from default keychain")
```
alfredodeza, line 71: since this is falling back, it would be useful to have the log message say so. Otherwise, with debug enabled, it isn't exactly clear why the keychain is used unless one is reading the source

pkg/image/source.go
wagoodman: this is specific to the docker daemon, not directly to a registry --to answer your question, no, docker should always return a version if you get a ping cmd response. (context, the reason why this was changed was to allow for registry to be a fallback in case the docker daemon wasn't available... however, in order to determine that you need to reach out to the docker daemon... the "simplest" way to reach out with little consequence was with a ping command)

pkg/image/oci/registry_provider.go
wagoodman: Without `TLS` in the name it would be hard to reason about what verification is being skipped. Take for example the stdlib `tls.Config`:
```golang
tls.Config{InsecureSkipVerify: true}
```
in this example it's relatively clear that the verification being skipped somehow relates to TLS, whereas for the registry options it would not be:
```golang
image.RegistryOptions{InsecureSkipVerify: true}
```
this could be misconstrued as skipping some sort of verification relating to the image or the registry, but it's difficult to pin down that it's about TLS here. 

I changed the config option to be more in line with how'd you'd configure `kubectl` which takes a `--insecure-skip-tls-verify` option for similar behavior.

pkg/image/registry_options.go
wagoodman: ahh yeah, I consider that approach a "guard clause", I can switch to that 👍

pkg/image/registry_options.go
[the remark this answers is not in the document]
wagoodman: good question, we could mimic docker more closely in this way --where they allow TLS verification configuration on a per-registry basis.

pkg/image/source.go
wagoodman: that's a good point, I'll bump this to, say, 10?

pkg/image/registry_options.go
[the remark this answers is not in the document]
wagoodman: After a little more thought, I think it makes sense to leave this as is for now... but we can add it later as an enhancement since it isn't mutually exclusive with a global option.

pkg/image/source.go
alfredodeza: I think bumping this makes sense, but I call these magical numbers because we are guessing and not allowing other values to be set. We can go with 10, but I think we need to revisit this with a separate issue to make it configurable.

alfredodeza (approved):
This is good to go I think. Would like to see a follow-up for the timeout value for registry verification as an issue.

wagoodman:
righto @alfredodeza , we can revisit the timeout value for talking to the docker daemon (re: https://github.com/anchore/stereoscope/pull/64#discussion_r611905521)

Issue #68, opened 2021-05-21
Cannot authenticate to registry without specifying a specific "authority" config value


luhring:
**Background:** In order to ensure stereoscope isn't sending a given set of credentials to a container registry other than that which the user intends, stereoscope takes an [`Authority`](https://github.com/anchore/stereoscope/blob/d577f30b19e64b4c4b239bd2f23403014fa612fd/pkg/image/registry_options.go#L17) config value. This is intended to prevent, for example, sending the user's Docker Hub credentials to the Amazon ECR service. The way this works is that stereoscope compares the given registry host value (e.g. `111122223333.dkr.ecr.us-west-2.amazonaws.com`) and ensures it matches the given credential set's `Authority` value before making use of the credentials.

The registry value is sometimes made explicit by the image string (e.g. `111122223333.dkr.ecr.us-west-2.amazonaws.com/my-web-app:latest`) and sometimes left implicit (e.g. `ubuntu:latest`) in which case the registry value is determined automatically by the go-containerregistry library.

**Problem:** If the user needs to authenticate with a registry, and they don't provide the correct `Authority` config value that corresponds to the image string they've provided (which includes the case of not providing an `Authority` value at all), authentication will fail because stereoscope will not use the provided credentials.

**Requirement:** Stereoscope shouldn't attempt to compare the registry host value with the `Authority` value in the case where the `Authority` value is an empty string. Such a case indicates that the user is expecting the credentials they provide to be applied to the registry determined by inspecting their provided image string. For example, if they are specifying `registry:registry.gitlab.com/my-org/my-project:latest`, the user expects the provided credentials (e.g. a username and password pair) to be sent to the registry `registry.gitlab.com...`).

**Notes to developer:**

A change will be needed here: https://github.com/anchore/stereoscope/blob/main/pkg/image/registry_options.go#L27
As well as in downstream consumer config processing, e.g. here: https://github.com/anchore/grype/blob/main/internal/config/registry.go#L46

PR #69, opened 2021-05-23
Allow for empty authority value and refactor

luhring:
Addresses #68 

- Allows for empty `Authority` value for registry credentials, which will be usable with any registry
- Breaks out individual "credentials" concerns from "registry options" concerns (i.e. a collection of credentials)
- Refactors authenticator-creation logic and adds tests for new code paths

### DIFF
--- a/pkg/image/registry_credentials_test.go
+++ b/pkg/image/registry_credentials_test.go
@@ -83,23 +83,31 @@ func TestRegistryCredentials_Authenticator(t *testing.T) {
 
 func basicAuth(expected authn.Basic) func(*testing.T, authn.Authenticator) {
 	return func(t *testing.T, actual authn.Authenticator) {
+		t.Helper()
+
 		assertBasicAuth(t, expected, actual)
 	}
 }
 
 func bearerToken(expected authn.Bearer) func(*testing.T, authn.Authenticator) {
 	return func(t *testing.T, actual authn.Authenticator) {
+		t.Helper()
+
 		assertBearerToken(t, expected, actual)
 	}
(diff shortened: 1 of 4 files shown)


pkg/image/registry_credentials_test.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+func nilAuthenticator() func(*testing.T, authn.Authenticator) {
+	return func(t *testing.T, actual authn.Authenticator) {
+		assert.Nil(t, actual)
+	}
+}
+
+func assertBasicAuth(t *testing.T, expected authn.Basic, actual authn.Authenticator) {
```
alfredodeza, line 102: why these don't use `t.Helper()` ?

wagoodman (approved):
nice improvements 💯

pkg/image/registry_credentials_test.go
luhring: @alfredodeza Because I forgot! Awesome catch!

Calling `t.Helper()` would be nice here (and also in the corresponding assertion functions), because it makes the line reported to the user this: `test.authenticatorAssertion(t, actual)`, which is in the actual test code.

(The full error trace is still reported regardless of calls to `t.Helper()`, so it's easy to see where in the helper functions this bubbled up from.)

I shall update this!

luhring: Updated here: https://github.com/anchore/stereoscope/pull/69/commits/3deea0cfd6231293c04d5028cf9324f32297b790

PR #70, opened 2021-06-26
Add warning log level

wagoodman:
There was existing code using a warning log level from `apex/log` which is incorrect. This PR adds the warning level to the package logger and fixes the import.


dakaneye (approved):
LGTM

PR #71, opened 2021-06-27
Add per-image scoped temp dir cleanup functions

wagoodman:
This PR changes the tempfile generation to share the same parent directory for the same call to `GetImage` or `GetImageFromSource`. This change requires that cleanup functions are exposed on these `GetImage*` calls to cleanup temp dirs created by that particular call. 

Today the temp file generation is globally handled with the `Cleanup` function, which (correctly) breaks when used as a test cleanup function, since it affects global state. Test cleanup calls now leverage the more narrowly-scoped cleanup function returned from `GetImage*` functions. 

This PR also opts to leave the global `Cleanup` function in tact. Why? Callers (such as syft and grype) may need to interrupt the process and exit, but also make certain that any temp directories are cleaned up. This PR updates the `file.TempDirGenerator` tracks all temp generators created and temp directories created from a root object such that the root object can cleanup all temp directories from all nested generators. This allows callers to have the "clean everything up and exit" option for exceptional cases (today, syft needs this).

Specific changes:
- Removes locking and individual dir creation tracking from `file.TempDirGenerator`
- A single `TempDirGenerator` now keeps all created temp dirs within a shared parent dir.
- Adds the ability to create additional `TempDirGenerators` from a `TempDirGenerator`, which will create a sibling temp dir with the same prefix as the parent prefix. All directories created from the child generator are persisted in the sibling temp dir from the parent generator.
- `TempDirGenerator.Cleanup` calls cleanup on all child generators as well as removing the shared top directory among all created temp dirs for the generator.
- adds `image.Image.Cleanup()` deletes the temp dir scoped for that particular image.

Partially addresses https://github.com/anchore/syft/issues/416 
Fixes https://github.com/anchore/stereoscope/issues/96

### DIFF
--- a/pkg/file/temp_dir_generator.go
+++ b/pkg/file/temp_dir_generator.go
@@ -1,50 +1,65 @@
 package file
 
 import (
-	"fmt"
-	"io/ioutil"
 	"os"
-	"sync"
+	"strings"
 
 	"github.com/hashicorp/go-multierror"
 )
 
 type TempDirGenerator struct {
-	tempDir []string
-	lock    *sync.Mutex
+	rootPrefix   string
+	rootLocation string
+	generators   []*TempDirGenerator
 }
 
-func NewTempDirGenerator() TempDirGenerator {
-	return TempDirGenerator{
-		tempDir: make([]string, 0),
-		lock:    &sync.Mutex{},
+func NewTempDirGenerator(name string) *TempDirGenerator {
+	return &TempDirGenerator{
+		rootPrefix: name,
 	}
 }
 
-// NewTempDir creates an empty dir in the platform temp dir
-func (t *TempDirGenerator) NewTempDir() (string, error) {
-	t.lock.Lock()
-	defer t.lock.Unlock()
-
-	dir, err := ioutil.TempDir("", "stereoscope-cache")
-	if err != nil {
-		return "", fmt.Errorf("could not create temp dir: %w", err)
-	}
-
-	t.tempDir = append(t.tempDir, dir)
-	return dir, nil
-}
-
-func (t *TempDirGenerator) Cleanup() error {
-	t.lock.Lock()
-	defer t.lock.Unlock()
-
-	var allErrors error
-	for _, dir := range t.tempDir {
-		err := os.RemoveAll(dir)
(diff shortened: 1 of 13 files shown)


luhring (changes requested):
You've made some great improvements made here! Having the image creation provide its own cleanup function is **excellent** because it lets image cleanup be handled directly alongside a caller's use of the image. This makes it much easier for consumers to manage the side effects had by image creation operations.

(Quick aside: One alternative approach to consider would be to have the image's cleanup function be a method on the image itself. This way, any consumer of the image is capable of cleaning up the image's backing data (similar to how database connection objects frequently have a `conn.Close()` method).

Primary concern: We need to avoid global state wherever possible, since it creates opportunities for bugs to arise in the future and it typically complicates the codebase. Fortunately, because of the cleanup functions being associated with image creation, we have an opportunity to associate an image's temp data with the image object directly. This would eliminate the need to maintain a global store of temp directories, a maintenance that has gotten a bit more complicated in the current draft.

Image handling would become more atomic, and a consumer of any given image wouldn't need to know about anything more global than the image itself.

wagoodman:
The leveraging of global state is unfortunate, but also an existing behavior today. This PR has made a step in removing that global state (with the cleanup functions). I think we need to sync up on context on Monday as some perspectives have been over simplified.

wagoodman:
We're deferring this work for now

pkg/file/temp_dir_generator.go
```diff
@@ -1,50 +1,65 @@
...
 
 	"github.com/hashicorp/go-multierror"
 )
 
 type TempDirGenerator struct {
-	tempDir []string
-	lock    *sync.Mutex
+	rootPrefix   string
+	rootLocation string
+	generators   []*TempDirGenerator
```
luhring, line 13: nit: Consider renaming to something like `childGenerators` to make it more clear what the relationship is with the parent struct and why this field exists

luhring (approved):
Nice!

pkg/file/temp_dir_generator.go
wagoodman: renamed to `children` 👍

PR #73, opened 2021-08-13
add insecure registry(http) support

derekhjray:
add insecure registry support, such as 'http://...'

### DIFF
--- a/pkg/image/oci/registry_provider.go
+++ b/pkg/image/oci/registry_provider.go
@@ -38,7 +38,12 @@ func (p *RegistryImageProvider) Provide() (*image.Image, error) {
 		return nil, err
 	}
 
-	ref, err := name.ParseReference(p.imageStr)
+	options := make([]name.Option, 0, 2)
+	if p.registryOptions.InsecureSkipTLSVerify {
+		options = append(options, name.Insecure)
+	}
+
+	ref, err := name.ParseReference(p.imageStr, options...)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/image/oci/registry_provider.go
```diff
@@ -38,7 +38,12 @@ func (p *RegistryImageProvider) Provide() (*image.Image, error) {
 		return nil, err
 	}
 
-	ref, err := name.ParseReference(p.imageStr)
+	options := make([]name.Option, 0, 2)
+	if p.registryOptions.InsecureSkipTLSVerify {
```
kzantow, line 42: I was looking in to: https://github.com/anchore/grype/issues/334 and came across this PR.

This change will cause the [registry to be accessed using `http://`](https://github.com/google/go-containerregistry/blob/7e30746dde2f5a19b609f885dfef4e7b1d1503ad/pkg/name/registry.go#L76), which is not the same thing as [skipping TLS verification](https://github.com/anchore/stereoscope/blob/d577f30b19e64b4c4b239bd2f23403014fa612fd/pkg/image/oci/registry_provider.go#L74) I think a more appropriate change would be to add another option specific for accessing the registry using `http://`.

kzantow:
I've merged an option for http support here: https://github.com/anchore/stereoscope/pull/75 is it okay to close this PR?

derekhjray:
Thank you, it's ok to close this PR
Is it possible to modify the stereoscope version syft that depends this package?

PR #75, opened 2021-08-13
Add option to allow registry connections over http (supersedes #73)

kzantow:
Updated #73 to use another option, because TLS validation and http/https are different concerns.

NOTE: this is a prerequisite of https://github.com/anchore/grype/issues/334

### DIFF
--- a/pkg/image/oci/registry_provider.go
+++ b/pkg/image/oci/registry_provider.go
@@ -38,7 +38,12 @@ func (p *RegistryImageProvider) Provide() (*image.Image, error) {
 		return nil, err
 	}
 
-	ref, err := name.ParseReference(p.imageStr)
+	options := make([]name.Option, 0, 2)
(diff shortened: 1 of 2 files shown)


pkg/image/oci/registry_provider.go
```diff
@@ -38,7 +38,12 @@ func (p *RegistryImageProvider) Provide() (*image.Image, error) {
 		return nil, err
 	}
 
-	ref, err := name.ParseReference(p.imageStr)
+	options := make([]name.Option, 0, 2)
```
luhring, line 41: nit: This line declares a new variable (`options`) whose value is a slice, specifically with an allocated capacity of `2`, and a length of `0`. Then below, conditionally, the slice length grows from 0 items to 1 item. To me, it's unclear why the capacity of 2 was chosen if the largest the slice's length ever becomes is 1.

luhring (approved):

PR #76, opened 2021-08-16
chore: refactor prepareReferenceOptions and tests

kzantow:
Follow-up to #75 to add a basic unit test

### DIFF
--- a/pkg/image/oci/registry_provider.go
+++ b/pkg/image/oci/registry_provider.go
@@ -38,12 +38,7 @@ func (p *RegistryImageProvider) Provide() (*image.Image, error) {
 		return nil, err
 	}
 
-	var options []name.Option
-	if p.registryOptions.InsecureUseHTTP {
-		options = append(options, name.Insecure)
-	}
-
-	ref, err := name.ParseReference(p.imageStr, options...)
+	ref, err := name.ParseReference(p.imageStr, prepareReferenceOptions(p.registryOptions)...)
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse registry reference=%q: %+v", p.imageStr, err)
 	}
@@ -72,6 +67,14 @@ func (p *RegistryImageProvider) Provide() (*image.Image, error) {
 	}
 
 	return image.NewImage(img, imageTempDir, metadata...), nil
+}
+
+func prepareReferenceOptions(registryOptions *image.RegistryOptions) []name.Option {
+	var options []name.Option
+	if registryOptions.InsecureUseHTTP {
+		options = append(options, name.Insecure)
+	}
+	return options
 }
 
 func prepareRemoteOptions(ref name.Reference, registryOptions *image.RegistryOptions) []remote.Option {
--- a/pkg/image/oci/registry_provider.go
+++ b/pkg/image/oci/registry_provider.go
@@ -71,7 +71,7 @@ func (p *RegistryImageProvider) Provide() (*image.Image, error) {
 
 func prepareReferenceOptions(registryOptions *image.RegistryOptions) []name.Option {
 	var options []name.Option
-	if registryOptions.InsecureUseHTTP {
+	if registryOptions != nil && registryOptions.InsecureUseHTTP {
 		options = append(options, name.Insecure)
 	}
 	return options
--- a/pkg/image/oci/registry_provider_test.go
+++ b/pkg/image/oci/registry_provider_test.go
@@ -1,0 +1,46 @@
+package oci
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/anchore/stereoscope/pkg/image"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_prepareReferenceOptions(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    image.RegistryOptions
+		expected []name.Option
+	}{
+		{
+			name:     "not InsecureUseHTTP",
+			input:    image.RegistryOptions{},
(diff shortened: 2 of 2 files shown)


pkg/image/oci/registry_provider.go
```diff
@@ -38,12 +38,7 @@ func (p *RegistryImageProvider) Provide() (*image.Image, error) {
 		return nil, err
 	}
 
-	var options []name.Option
-	if p.registryOptions.InsecureUseHTTP {
-		options = append(options, name.Insecure)
-	}
-
-	ref, err := name.ParseReference(p.imageStr, options...)
+	ref, err := name.ParseReference(p.imageStr, prepareReferenceOptions(p.registryOptions)...)
```
luhring, line 41: Because `p.registryOptions` is a pointer type, we should check to see if its value is `nil` at some point. In this situation, I could see two choices:

1. Check for its nilness here in the body of `Provide`. If you took this approach, you could simplify the signature of `prepareReferenceOptions` by having it accept a `image.RegistryOptions` (which can be relied on as not nil) instead of a `*image.RegistryOptions`.
2. Put the nilness check inside the body of `prepareReferenceOptions`.

The reason this matters is that `prepareReferenceOptions` has the expression `registryOptions.InsecureUseHTTP` which will assume that `registryOptions` **_is not nil_**, and if it actually is nil, the application will panic.

pkg/image/oci/registry_provider_test.go
```diff
@@ -0,0 +1,46 @@
...
+				InsecureUseHTTP: true,
+			},
+			expected: []name.Option{name.Insecure},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			out := prepareReferenceOptions(&test.input)
+			assert.Equal(t, len(test.expected), len(out))
```
luhring, line 35: Optional — in case you're interested, the `assert` package has a native length assertion. (But there's nothing wrong with the way you've written this, either.)

```suggestion
			assert.Len(t, out, len(test.expected))
```

pkg/image/oci/registry_provider_test.go
```diff
@@ -0,0 +1,46 @@
...
+			if test.expected == nil {
+				assert.Equal(t, test.expected, out)
+			} else {
```
luhring, line 38: Question: Just double-checking my understanding of the intent here — in addition to ensuring the slice length is correct, we're adding this nil-case check for the scenario where both slices have a length of zero, but one of the slices is non-nil?

pkg/image/oci/registry_provider_test.go
```diff
@@ -0,0 +1,46 @@
...
+				// cannot compare functions directly
```
luhring, line 39: TIL! 😄

luhring (commented):
Looks good overall! Added some suggestions and questions.

pkg/image/oci/registry_provider_test.go
kzantow: I typically prefer coming up with an equals check in tests instead of relying on some behavior defined in a library helper method, I think it makes things easier for everyone to follow who may not be familiar with said testing library, but this is just my preference.

pkg/image/oci/registry_provider_test.go
kzantow: correct, maybe I misunderstood... I have `var options []name.Option`, which is nil until it's later appended to `options = append(options, name.Insecure)` ? so this function returns `nil` or a 1-option array with the `Insecure` flag

pkg/image/oci/registry_provider.go
kzantow: Added `nil` check to `prepareReferenceOptions`: https://github.com/anchore/stereoscope/pull/76/files#diff-a48bf3359f783f78022ab9cd098b21e3c4c9262c2e6319f55a11bc0004cddc09R74

pkg/image/oci/registry_provider_test.go
luhring: Yes, that's all correct. 🎉 

Maybe I could rephrase my prior comment to be more clear: We're already testing to ensure the length is the same, just a few lines up. So if we expect a zero-item slice (`nil`), and we get a zero-item slice (`nil`), that assertion will succeed.

So my question is: what is _this assertion_ checking for that hasn't already been checked?

```go
			if test.expected == nil {
				assert.Equal(t, test.expected, out)
			}
```

In a way, this is a different way of asserting something we've already asserted. But there's an exception to that statement, a particular case where that's not true. Zero-length slices can be `nil`, or they can be allocated (i.e. not `nil`) but have no array items. This assertion would alert us to a scenario where it was important to us that the function returns specifically the `nil` kind of zero-item slice, but the function instead returned an allocated zero-item slice.

I'm sort of speculating about what the intent is here, so what I'm saying is a bit contrived. In short, I think what you have is great and doesn't require a change. I'm just trying to follow along as I'm reading. 😄

luhring (approved):
Nice refactoring!

pkg/image/oci/registry_provider_test.go
kzantow: Yeah, just basically testing that we're returning `nil` as expected instead of an allocated 0-length array, also I think the `array[0]` would panic? maybe not

luhring: Cool, thanks for the clarification.

> I think the array[0] would panic

Yep, it would, in either of the scenarios... Using a nil slice would cause a `nil pointer dereference` panic, and accessing index `0` of a non-`nil` slice with _zero items_ would cause an `index out of range` panic.

Issue #7, opened 2020-05-09
Add podman support


wagoodman:
- Add `podman` image source, enabling `podman:<image>` parsing
- Add `podman` image provider, enabling tarball extraction from the podman ~via it's varlink api~

jonasagx:
podman's libpod uses CGO: https://github.com/containers/podman/blob/main/libpod/lock/shm/shm_lock.go#L5

jonasagx:
Another way to support podman is through a docker client. Podman is docker-compatible making docker podman-compatible :)  

A docker client can use podman to work, however it needs an ssh tunnel. Syft looks for `DOCKER_HOST` while setting its docker client.

The process looks like: Syft --> Docker client --> podman daemon --> image pull.

Proof of dockerCLI<-->podman compatibility:
```sh
// start podman daemon AND make sure docker daemon is not running
$ podman machine start

// check its connection port
$ podman system connection ls
Name                         Identity                                  URI
podman-machine-default*      /Users/user/.ssh/podman-machine-default  ssh://core@localhost:[PORT]/run/user/1000/podman/podman.sock
podman-machine-default-root  /Users/user/.ssh/podman-machine-default  ssh://root@localhost:[PORT]/run/podman/podman.sock

// make a tunnel to podman socket
$ ssh -nNT -L/tmp/podman.sock:/run/user/1000/podman/podman.sock -i ~/.ssh/podman-machine-default ssh://core@localhost:[PORT]

// on another terminal 
$ export DOCKER_HOST=unix:///tmp/podman.sock
$ docker version // server should be a podman engine

// explicitly tell syft to use docker daemon 
$ syft docker:alpine:latest -vv
...
[0000] DEBUG image: source=DockerDaemon location=alpine:latest from-lib=stereoscope
...
```

I am not sure why the tunnel is necessary since the socket file is in the same file system of the host. Also dropping the tunnel reduces complexity.

jonasagx:
![podman-ssh-tunnel](https://user-images.githubusercontent.com/774704/140463676-9d16b907-51df-488b-bf34-2534bab55982.png)[1]

Podman uses a tunnel to handle communication with its daemon. That is why the tunnel above was necessary, and why for stereoscope to talk to podman we should have a param to pass an ssh key.

[1] - https://www.redhat.com/sysadmin/podman-clients-macos-windows

jonasagx:
Docker client has an option to customize the ssh call made during each request:
```golang
[...]

PR #87, opened 2021-11-29
Add support for Podman daemon

jonasagx:
Adds support to podman via docker client. For now the main difference between docker's and podman's client is how they communicate to their daemons. This PR adds a ssh bastion to docker's http client so it can dial http requests to podman. 

closes https://github.com/anchore/stereoscope/issues/7


jonasagx:
integration tests pass when I run them locally, but on github they fail without an error message:
```
--- FAIL: TestSimpleImage (4.17s)
    --- PASS: TestSimpleImage/FromTarball (2.10s)
    --- FAIL: TestSimpleImage/FromDocker (0.05s)
    --- FAIL: TestSimpleImage/FromPodman (0.05s)
    --- PASS: TestSimpleImage/FromOciTarball (1.91s)
    --- PASS: TestSimpleImage/FromOciDirectory (0.06s)
```

Is there another place to look for logs?

wagoodman:
@jonasagx I've updated the validations pipeline here https://github.com/anchore/stereoscope/pull/88 ... once we get that PR in and rebase here we can troubleshoot further. Why this way? I noticed that the failure is between the go versions in the test matrix, which we've removed (it was deemed as unnecessary and there was a bug we squashed that looks similar to this behavior)

internal/docker/podman.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/anchore/stereoscope/internal/log"
+	"github.com/docker/docker/client"
+	"github.com/docker/docker/pkg/homedir"
+	"github.com/pkg/errors"
+	"golang.org/x/crypto/ssh"
+)
+
+var (
+	makePodmanClientOnce sync.Once
+	defaultHost          = "ssh://core@localhost:63753/run/user/1000/podman/podman.sock?secure=false"
```
wagoodman, line 23: this has what appears to be an ephemeral port (which will change frequently)

internal/docker/podman.go
```diff
@@ -0,0 +1,171 @@
...
+)
+
+func getPodmanClient() (*client.Client, error) {
+	makePodmanClientOnce.Do(func() {
```
wagoodman, line 27: this function presumes that podman will only be available over SSH, which is true on a mac, however, on a linux box this is not true.

pkg/image/source.go
```diff
@@ -35,6 +37,7 @@ var sourceStr = [...]string{
 	"OciDirectory",
 	"OciTarball",
 	"OciRegistry",
+	"PodmanDaemon",
```
spiffcs, line 40: This is what I was referencing in the above comment about sharing the value for when we parse the type of Daemon:
`internal/docker/client.go`

pkg/image/source_test.go
```diff
@@ -2,13 +2,14 @@ package image
...
-	"github.com/mitchellh/go-homedir"
-	"github.com/spf13/afero"
 	"io"
 	"os"
 	"path"
 	"strings"
 	"testing"
+
+	"github.com/mitchellh/go-homedir"
+	"github.com/spf13/afero"
```
spiffcs, line 12: Nice - love seeing the import tools run

spiffcs (commented):
Review comments added surrounding the SSH package as well as some smaller comments.

internal/docker/client.go
```diff
@@ -26,7 +31,11 @@ func GetClient() (*client.Client, error) {
 		host := os.Getenv("DOCKER_HOST")
 
 		if strings.HasPrefix(host, "ssh") {
-			helper, err := connhelper.GetConnectionHelper(host)
+			var helper *connhelper.ConnectionHelper
```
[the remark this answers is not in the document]
jonasagx: `var ()` looks cleaner and less repetitive

internal/docker/podman.go
```diff
@@ -0,0 +1,171 @@
...
+			log.Debugf("using $CONTAINER_HOST: %s", v)
+			host = v
+		}
+
+		if v, found := os.LookupEnv("PODMAN_HOST"); found && v != "" {
+			log.Debugf("using $PODMAN_HOST: %s", v)
+			host = v
+		}
+
+		_url, err := url.Parse(host)
```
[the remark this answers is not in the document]
jonasagx: yes, `hostURL` looks better

internal/docker/podman.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return &http.Client{
+		Transport: &http.Transport{
+			DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
+				return bastion.Dial("unix", _url.Path)
+			},
+		}}, nil
+}
+
+func publicKey(path string, passphrase []byte) (ssh.Signer, error) {
```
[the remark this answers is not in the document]
jonasagx: for the equivalent operation the lib recommends an implementation similar to this PR: https://pkg.go.dev/golang.org/x/crypto/ssh#example-PublicKeys

jonasagx:
FYI: The implementation this far doesn't work on a linux host (using a plain Podman installation)

client.go
```diff
@@ -38,6 +40,7 @@ func GetImageFromSource(imgStr string, source image.Source, registryOptions *ima
 
 	img, err := provider.Provide()
 	if err != nil {
+		log.Errorf("failed to use provider: %v", err)
```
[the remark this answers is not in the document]
wagoodman: this is a log statement, so `%w` can't be used. However, this should probably be adjusted to just returning an error (without a log call):
```golang
	if err != nil {
		return nil, fmt.Errorf("failed to use provider: %w", err)
        }
```

examples/basic.go
```diff
@@ -2,18 +2,23 @@ package main
...
 	"github.com/anchore/stereoscope/pkg/file"
+	"github.com/anchore/stereoscope/pkg/filetree/filenode"
+	"github.com/sirupsen/logrus"
 )
 
 func main() {
 	// note: we are writing out temp files which should be cleaned up after you're done with the image object
 	defer stereoscope.Cleanup()
 
+	lctx := logrus.New()
```
wagoodman, line 18: nice ❤️

pkg/image/source.go
```diff
@@ -125,6 +131,7 @@ func detectSource(fs afero.Fs, userInput string) (Source, string, error) {
 					// the docker daemon exists and is accessible
 					return DockerDaemonSource, userInput, nil
 				}
+				log.Errorf("failed to ping docker daemon: %v", err)
```
wagoodman, line 134: our latest convention is that `log.Error` is reserved for when something goes wrong that is not recoverable (results in a premature exit). In cases like this where we continue with normal execution we should use `log.Warn*`.

pkg/image/source.go
```diff
@@ -125,6 +131,7 @@ func detectSource(fs afero.Fs, userInput string) (Source, string, error) {
 					// the docker daemon exists and is accessible
 					return DockerDaemonSource, userInput, nil
```
wagoodman, line 132: This entire `case UnknownSource: ...` should probably get updated to ping podman if docker can't be found and fallback to returning `PodmanDaemonSource`.

pkg/image/source.go
```diff
@@ -43,6 +46,7 @@ var AllSources = []Source{
 	OciDirectorySource,
 	OciTarballSource,
 	OciRegistrySource,
+	PodmanDaemonSource,
```
wagoodman, line 49: nit: I dunno if I believe in my next comment, so lots of grains of salt here... Since a popular podman installation configuration is socket-activation (and not a long running daemon) then should this be named more simply `PodmanSource`?

abitrolly:
Is it possible to integrate with `podman` without using the daemon? Because the daemon needs to be started manually.

wagoodman:
@abitrolly We would like to though I don't see an easy path forward. Specifically, the toolchain that depends on stereoscope requires that CGO is disabled (only allow pure-go dependencies). Doing this means we can keep our tools as static binaries that anyone can curl and run without needing to install other packages. More investigation is needed here (for instance https://github.com/containers/podman/blob/main/Makefile#L154-L155).

This PR will at least get us started with podman support. If you have podman installed and you're on a linux box I think the only step you'd need to take to leverage the work done in this PR is:
```
systemctl --user enable --now podman.socket
```

internal/docker/podman.go
```diff
@@ -0,0 +1,236 @@
...
+)
+
+var (
+	// Podman supports the following connections
+	// rootless "unix://run/user/$UID/podman/podman.sock" (Default)
+	// rootfull "unix://run/podman/podman.sock (Default)
+	// remote rootless ssh://engineering.lab.company.com/run/user/$UID/podman/podman.sock
+	// remote rootfull ssh://root@10.10.1.136:22/run/podman/podman.sock
+
+	localRootlessPath     = getLocalRootlesPath()
```
wagoodman, line 29: this doesn't need to be a variable in static space anymore

internal/docker/podman.go
```diff
@@ -0,0 +1,236 @@
...
+	defaultRemoteRootless = fmt.Sprintf("ssh://core@localhost:63753%s?secure=false", localRootlessPath)
+	defaultLocalRootless  = fmt.Sprintf("unix://%s", localRootlessPath)
```
wagoodman, line 31: this is unused and could probably be deleted

internal/docker/podman.go
```diff
@@ -0,0 +1,236 @@
...
+func getLocalRootlesPath() string {
+	return fmt.Sprintf("unix:///run/user/%d/podman/podman.sock", os.Getuid())
+}
+
+func podmanOverSSH() (*client.Client, error) {
+	log.Debug("using docker client to connect to podman over ssh")
+	var clientOpts = []client.Opt{
+		client.WithAPIVersionNegotiation(),
+		// TODO: is this needed?
+		client.FromEnv,
```
wagoodman, line 43: will this mess up settings based on `DOCKER_HOST`? (should we not respond to docker-specific args?)

internal/docker/podman.go
```diff
@@ -0,0 +1,236 @@
...
+
+	return dockerClient, err
+}
+
+func podmanViaUnixSocket() (*client.Client, error) {
+	log.Debug("using docker client to connect to podman over local unix socket")
+	var clientOpts = []client.Opt{
+		client.WithAPIVersionNegotiation(),
+		// TODO: is this needed?
+		client.FromEnv,
```
wagoodman, line 89: same comment https://github.com/anchore/stereoscope/pull/87/files#r764953646

internal/docker/podman.go
```diff
@@ -0,0 +1,236 @@
...
+
+	httpClient, err := makeClient(hostURL)
+	if err != nil {
+		return nil, fmt.Errorf("making http client: %w", err)
+	}
+
+	clientOpts = append(clientOpts, func(c *client.Client) error {
+		return client.WithHTTPClient(httpClient)(c)
+	})
+	clientOpts = append(clientOpts, client.WithHost("http://d"))
```
wagoodman, line 74: this is probably worth a healthy comment on the need for `http://d`

internal/docker/podman.go
```diff
@@ -0,0 +1,236 @@
...
+
+	if v, found := os.LookupEnv("CONTAINER_HOST"); found && v != "" {
+		log.Debugf("using $CONTAINER_HOST: %s", v)
+		host = v
+	}
+
+	if v, found := os.LookupEnv("PODMAN_HOST"); found && v != "" {
+		log.Debugf("using $PODMAN_HOST: %s", v)
+		host = v
+	}
```
wagoodman, line 57: this seems like a shared concern that could be broken out between the two functions that return clients (this was 100% my doing while we were troubleshooting last night, but didn't want to loose the thought)

internal/docker/podman.go
```diff
@@ -0,0 +1,236 @@
...
+
+	dockerClient, err := client.NewClientWithOpts(clientOpts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed create local client for podman: %w", err)
+	}
+
+	return dockerClient, err
+}
+
+func GetClientForPodman() (*client.Client, error) {
```
wagoodman, line 109: this detection could probably be based on the existence of a podman socket at a predetermined path instead of by only OS since podman over SSH is a valid setup on linux. Maybe a hybrid approach where OS is used for windows/darwin, but the linux path has an extra conditional?

internal/docker/podman.go
```diff
@@ -0,0 +1,236 @@
...
+func unixClient(hostURL *url.URL) (*http.Client, error) {
+	return &http.Client{
+		Transport: &http.Transport{
+			DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
+				return (&net.Dialer{}).DialContext(ctx, "unix", hostURL.Path)
+			},
+			DisableCompression: true,
+		},
+	}, nil
+}
```
wagoodman, line 132: These could be deleted now (I think it was me that didn't clean this up while we were troubleshooting)

internal/docker/podman.go
```diff
@@ -0,0 +1,236 @@
...
+	if v, found := os.LookupEnv("CONTAINER_SSHKEY"); found && len(identity) == 0 {
+		log.Debugf("using $CONTAINER_SSHKEY: %s", v)
+		return v
+	}
+
+	return identity
+}
+
+// NOTE: code inspired by Podman's client: https://github.com/containers/podman/blob/main/pkg/bindings/connection.go#L177
+func sshClient(hostURL *url.URL) (*http.Client, error) {
```
wagoodman, line 145: nit: this function could probably be broken up to improve grokability

abitrolly:
@wagoodman is it possible to just use CLI `podman` to query and fetch images? I don't know if there is any official CLI API it provides. I guess these are the calls that are needed.

* [ ] Find if the image or container exists by ID or name
* [ ] Fetch the image or container if it is not present
* [ ] Read the image or container data

internal/docker/podman.go
```diff
@@ -0,0 +1,273 @@
...
+
+var (
+	// Podman supports the following connections
+	// rootless "unix://run/user/$UID/podman/podman.sock" (Default)
+	// rootfull "unix://run/podman/podman.sock (Default)
+	// remote rootless ssh://engineering.lab.company.com/run/user/$UID/podman/podman.sock
+	// remote rootfull ssh://root@10.10.1.136:22/run/podman/podman.sock
+
+	localRootlessPathTemplate = "/run/user/%d/podman/podman.sock"
+	defaultRemoteRootless     = "ssh://core@localhost:63753%s?secure=false"
```
wagoodman, line 34: is `63753` a port number that is not expected to change from system to system? Is it driven by podman configuration at all?

wagoodman, line 34: `secure=false` seems like a potential security problem (though unclear). Can you elaborate on what specifically is insecure about this connection?

internal/docker/podman.go
```diff
@@ -0,0 +1,273 @@
...
+		return "", err
+	}
+
+	sock := fmt.Sprintf(localRootlessPathTemplate, uid)
+	return fmt.Sprintf(defaultRemoteRootless, sock), nil
+}
+
+func getRemoteUID() (int, error) {
+	// TODO: expose these ssh params to caller. End user should have settings for them
+	cmd := exec.Command("ssh", "-i", "~/.ssh/podman-machine-default", "-l", "core", "-p", "63753", "--", "localhost", "id -u")
```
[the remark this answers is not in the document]
abitrolly: `podman exec`?

```shell
➜  ~ podman exec -h
Run a process in a running container

Description:
  Execute the specified command inside a running container.


Usage:
  podman exec [options] CONTAINER [COMMAND [ARG...]]

Examples:
  podman exec -it ctrID ls
  podman exec -it -w /tmp myCtr pwd
  podman exec --user root ctrID ls
```

But why one need a remote UID at all?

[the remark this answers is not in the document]
jonasagx: Podman's cmd also offers `podman info -f "{{.Host.RemoteSocket.Path}}"`

@abitrolly: The UID is used in the socket path created by Podman: `ssh://core@localhost:63753/run/user/1000/podman/podman.sock` the 1000 here is the user ID. 
Side note: the root socket path does have a simpler address: `ssh://root@localhost:63753/run/podman/podman.sock` but AFAIK the port can be different, which is an issues this PR doesn't solve, yet.

**Using Podman's config file**
Podman's engine address is also in a config file: `~/.config/containers/containers.conf`, toml formatted. A config file seems cleaner to use and adjust, when compared with running a binary in some place in the file tree. We can also support different config files via cmd flags

Reading info from a config file is consistent with how we use docker's lib currently:
https://github.com/jonasagx/stereoscope/blob/main/internal/docker/client.go#L44-L49

internal/docker/podman.go
jonasagx: In Podman's docs and code I don't see info either it changes or not. Podman does offer a config file where we can get this info from. Vide comment below

jonasagx: I copy-pasta'ed it from Podman's client code, and disabled it until I understand it. This flag is meant to control whether the ssh handshake validates the server's host key against the local known keys in `.ssh/known_hosts`, which is important when talking to remote Podman servers.

internal/docker/podman.go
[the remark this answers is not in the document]
abitrolly: @jonasagx UID is the local host user, so why need SSH to access it? As described in https://docs.podman.io/en/latest/markdown/podman-system-service.1.html the socket path is unix://$XDG_RUNTIME_DIR/podman/podman.sock

internal/docker/podman.go
jonasagx: I changed it so the full address comes from Podman's config file

internal/docker/podman.go
jonasagx: I updated it so ssh and unix path are supported, which covers linux & Mac instances.

internal/docker/podman.go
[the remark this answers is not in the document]
jonasagx: @abitrolly I agree `$XDG_RUNTIME_DIR` can be used, thank you for the suggestion, but that doesn't work in all scenarios[1]. 
FYI: I dropped SSH use for checking UID, that info will come from `$XDG_RUNTIME_DIR` or Podman's config file.

[1] For windows and macOS where Podman engine runs on QEMU, SSH is necessary for client-server communication. Here `$XDG_RUNTIME_DIR` client and server might have different UID. More info at https://www.redhat.com/sysadmin/podman-clients-macos-windows

pkg/image/source.go
wagoodman: from conversation: I can see this both ways, I'll resolve this for now.

internal/docker/podman.go
jonasagx: resolved via config file

internal/docker/podman.go
```diff
@@ -0,0 +1,171 @@
...
+			return
+		}
+
+		instance = dockerClient
+	})
+
+	return instance, instanceErr
+}
+
+func sshClient(_url *url.URL, passPhrase string, identity string) (*http.Client, error) {
```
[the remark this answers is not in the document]
jonasagx: Sorry for the name confusion: this function was wrapping too much logic, it was broken down. In summary: it creates a http client based on ssh connections, which is an abstraction layer above `ssh.NewClient`.

internal/docker/podman.go
```diff
@@ -0,0 +1,236 @@
...
+
+var (
+	// Podman supports the following connections
+	// rootless "unix://run/user/$UID/podman/podman.sock" (Default)
+	// rootfull "unix://run/podman/podman.sock (Default)
+	// remote rootless ssh://engineering.lab.company.com/run/user/$UID/podman/podman.sock
+	// remote rootfull ssh://root@10.10.1.136:22/run/podman/podman.sock
+
+	localRootlessPath     = getLocalRootlesPath()
+	defaultRemoteRootless = fmt.Sprintf("ssh://core@localhost:63753%s?secure=false", localRootlessPath)
```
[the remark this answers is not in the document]
jonasagx: refactored, added comments about `secure` flag.

internal/podman/config.go
```diff
@@ -0,0 +1,170 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/url"
+	"path/filepath"
+
+	"github.com/docker/docker/pkg/homedir"
+	"github.com/pelletier/go-toml"
+)
+
+// env vars
```
wagoodman, line 13: old comments from dev that probably should be removed

internal/podman/config.go
```diff
@@ -0,0 +1,170 @@
...
+	if err := toml.Unmarshal(configBytes, &cc); err != nil {
+		return nil, err
+	}
+	return &cc, nil
+}
+
+var (
+	// _configPath is the path to the containers/containers.conf
+	// inside a given config directory.
+	_configPath = "containers/containers.conf"
```
wagoodman, line 133: - why prefix with `_`?
- This is better served as a `const`

internal/podman/config.go
```diff
@@ -0,0 +1,170 @@
...
+	// paths holds a list of config files, they are sorted from
+	// the least to the most relevant, i.e. a config file in
+	// the home directory has precedence over all other configs
+	configPaths = []string{
```
wagoodman, line 137: should be using `path.Join` for these string concatenations

internal/podman/config.go
```diff
@@ -0,0 +1,170 @@
...
+		// holds the default containers config path
+		"/usr/share/" + _configPath,
+		// holds the default config path overridden by the root user
+		"/etc/" + _configPath,
+		// holds the containers config path overridden by the rootless user
+		filepath.Join(homedir.Get(), "/.config/", _configPath),
+	}
+)
+
+func getUnixSocketAddress(paths []string) (address string) {
```
wagoodman, line 147: nit: could simplify this to something like
```golang
func getUnixSocketAddress(paths []string) string {
	for _, p := range paths {
		if a := findUnixAddressFromFile(p); a != "" {
			return a
		}
	}
        return ""
}
```

internal/podman/config.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func getSSHAddress(paths []string) (address, identity string) {
+	for _, p := range paths {
+		a, id := findSSHConnectionInfoFromFile(p)
+		// overwriting here is intentional, as a way to
+		// prioritize different config files
+		if a != "" && id != "" {
+			address = a
+			identity = id
```
wagoodman, line 165: this is bailing at the last match, though it seems like returning on the first match would be better

internal/podman/ssh_test.go
```diff
@@ -0,0 +1,232 @@
...
+	const (
+		sshAddress = "ssh://core@localhost:45983/run/user/1000/podman/podman.sock"
+		sshKeyPath = "/home/jonas/.ssh/podman-machine-default"
+	)
+
+	address, identity := getSSHAddress(paths)
+	assert.Equal(t, sshAddress, address)
+	assert.Equal(t, sshKeyPath, identity)
+
+	conf, err := newSSHConf(address, identity, "")
```
wagoodman, line 33: no request for change here, but you could do `cmp.Equal` to compare whole structs against expected structs, so you can do something like:
```golang
expected := mystruct{
  Value1: "thing"
  Value2: OtherStruct{
      Value3: "3",
   },
}

assert.True(t, cmp.Equal(expected, actual))

```

internal/podman/config.go
jonasagx: I agree on all the above, however Golang doesn't allow the use of filapath.Join to set constants, because go must know the const value during compilation time. So I am declaring it as a var with filepath.Join

internal/podman/config.go
jonasagx: ^ vide unix equivalent above: TLDR: that is intended behavior, I added tests around it

test/integration/podman_test.go
```diff
@@ -0,0 +1,134 @@
...
+		for scanner.Scan() {
+			t.Logf("%s: %s", label, scanner.Text())
+		}
+	}
+
+	show("out", stdout)
+	show("err", stderr)
+}
+
+// test conditions:
```
wagoodman, line 43: nit: I think this comment is out of date and should be deleted

test/integration/test-fixtures/podman/containers.conf
```diff
@@ -0,0 +1,20 @@
...
+  userns_size = 65536
+
+[engine]
+  num_locks = 2048
+  active_service = "ephemeral"
+  stop_timeout = 10
+  [engine.service_destinations]
+    [engine.service_destinations.ephemeral]
+      uri = "ssh://root@localhost:2222/run/podman/podman.sock"
+# TODO: dont hardcode this... this should be the volume mount from a running container
```
wagoodman, line 13: if this fixture still used? I think I added this for reference but it isn't used anywhere (so should be deleted)

wagoodman (approved):
great work 💯 ! Only left a few nits. From a standpoint of merging, this PR is ready, though it would be a good idea to get the Syft PR ready and working to show that no other changes are required before merging this in.

PR #91, opened 2021-12-15
Add file exclusion hook

kzantow:


test/integration/oci_registry_source_test.go
```diff
@@ -33,9 +33,6 @@ func TestOciRegistrySourceMetadata(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unable to get image: %+v", err)
 	}
-	if err := img.Read(); err != nil {
```
kzantow, line 36: Ultimately, `stereoscope.GetImage` above will call `img.Read` before returning and also return any error from the call

kzantow:
This was only needed for an alternate implementation of: https://github.com/anchore/syft/pull/695 but no longer

PR #92, opened 2021-12-19
Expose source in error message

luhring:
Just a few small tweaks, which helps out a lot on the consumer side, related to https://github.com/anchore/syft/issues/701. Will be accompanied by a Syft PR.

I also ran `go mod tidy` (I realized this isn't being enforced on `main` yet).


jonasagx (approved):
What if we rename this pr's tittle to explain its change in more details? Such as: expose source in error message. It makes for a more concise story told through release notes

luhring:
> What if we rename this pr's tittle to explain its change in more details? Such as: expose source in error message. It makes for a more concise story told through release notes

@jonasagx Great suggestion, thanks!

PR #93, opened 2021-12-21
Windows compatible development

spiffcs:
## Update stereoscope to be compatible with development on windows

Running tests on windows would prove difficult for the following reasons:
- bash shells called during testing for state configuration
  - (solution add build tags until future improvements can be made to orchestrate state correctly on each OS)
- skopeo does not have a windows release
  - (solution add build tags to ignore tests leveraging skopeo for windows until skopeo is supported) 
  - https://github.com/containers/skopeo/issues/715
- posix test paths were being incorrectly converted to windows paths causing mutliple tests to fail

Screen shots of behavior in order:
Bash Errors
Screenshot 2021-12-21 034531

Skopeo Errors
Screenshot 2021-12-21 034556

Inadvertant filepath conversion
Screenshot 2021-12-21 034504

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ RESET := $(shell tput -T linux sgr0)
 TITLE := $(BOLD)$(PURPLE)
 SUCCESS := $(BOLD)$(GREEN)
 # the quality gate lower threshold for unit test total % coverage (by function statements)
-COVERAGE_THRESHOLD := 45
+COVERAGE_THRESHOLD := 40
 
 ifeq "$(strip $(VERSION))" ""
     override VERSION = $(shell git describe --always --tags --dirty)
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ RESET := $(shell tput -T linux sgr0)
 TITLE := $(BOLD)$(PURPLE)
 SUCCESS := $(BOLD)$(GREEN)
 # the quality gate lower threshold for unit test total % coverage (by function statements)
-COVERAGE_THRESHOLD := 44
+COVERAGE_THRESHOLD := 48
 
 ifeq "$(strip $(VERSION))" ""
     override VERSION = $(shell git describe --always --tags --dirty)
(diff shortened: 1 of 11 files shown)


pkg/file/metadata_test.go
```diff
@@ -1,11 +1,14 @@
+// +build linux
```
kzantow, line 1: Should these also build mac?

pkg/filetree/glob.go
```diff
@@ -161,11 +187,6 @@ func (a *osAdapter) Open(name string) (fs.File, error) {
 	}, nil
 }
 
-// PathSeparator returns the standard separator between path entries for the underlying filesystem.
-func (a *osAdapter) PathSeparator() rune {
-	return []rune(file.DirSeparator)[0]
```
kzantow, line 166: 🎉

kzantow (approved):
LGTM, only real question is `// +build linux` should also include mac?

pkg/file/metadata_test.go
spiffcs: That's my fault - I'll update the tags to exclude windows rather than include all builds we want

spiffcs:
@kzantow excludes windows now rather than writing an allow list for all possible `GOOS`

Makefile
```diff
@@ -12,7 +12,7 @@ RESET := $(shell tput -T linux sgr0)
 TITLE := $(BOLD)$(PURPLE)
 SUCCESS := $(BOLD)$(GREEN)
 # the quality gate lower threshold for unit test total % coverage (by function statements)
-COVERAGE_THRESHOLD := 45
+COVERAGE_THRESHOLD := 48
```
kzantow, line 15: is there a specific reason for this change?

kzantow (approved):
this all looks fine to me & good addition to check for cycles!

wagoodman (approved):
I was hesitant to exclude some testing when running on a windows system at first, but after talking it through I think you're right. Over time we can improve our windows support in this area, not just overnight. Additionally in the future we can gain visibility into this and actively gate in our PRs by additionally running all CI checks on windows boxes. This is gonna be a marathon, not a sprint.

Thanks for the work identifying, carving out, and fixing these things 👍

Makefile
wagoodman: The current output from `make unit` is (from the latest run in the pipeline):
```
Coverage: 49.2
```
We try to bump the coverage threshold (if testing coverage falls below that threshold unit tests automatically fail) as we get more coverage (though this is a little close, but probably alright 👍 ).

PR #99, opened 2022-02-01
Propagate context though GetImage* functions

ajvpot:


wagoodman (approved):
Exposing context here for the number of underlying calls that already require this is a great idea --good call 🚀 I think a good rebase, DCO commits, and passing CI is all that's left 👍

PR #100, opened 2022-02-04
ensure progress tracker completes execution

jonasagx:
Fixes [#764](https://github.com/anchore/syft/issues/764)
Signed-off-by: Jonas Galvão Xavier <jonas.agx@gmail.com>

### DIFF
--- a/pkg/image/docker/daemon_provider.go
+++ b/pkg/image/docker/daemon_provider.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"math"
@@ -25,6 +26,10 @@ import (
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/wagoodman/go-partybus"
 	"github.com/wagoodman/go-progress"
+)
+
+var (
+	ErrEmptyImage = errors.New("cannot provide an empty image")
 )
 
 // DaemonImageProvider is a image.Provider capable of fetching and representing a docker image from the docker daemon API.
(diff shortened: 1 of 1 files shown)


pkg/image/docker/daemon_provider.go
```diff
@@ -193,7 +205,7 @@ func (p *DaemonImageProvider) Provide() (*image.Image, error) {
 		return nil, fmt.Errorf("unable to save image to tar: %w", err)
 	}
 	if nBytes == 0 {
-		return nil, fmt.Errorf("cannot provide an empty image")
+		return nil, ErrEmptyImage
```
kzantow, line 208: is there a specific reason the `ErrEmptyImage` var was introduced?

kzantow (approved):
This seems good to me, not being especially familiar with why the `copyProgress.SetComplete()` wasn't being called before

pkg/image/docker/daemon_provider.go
```diff
@@ -182,9 +197,6 @@ func (p *DaemonImageProvider) Provide() (*image.Image, error) {
 		}
 	}()
 
-	// cancel indeterminate progress
-	estimateSaveProgress.SetCompleted()
```
wagoodman, line 186: this should probably stick around. Why? The image save progress is only a guess (a timer counting up to a particular time where the overall progress would be considered at 50%). It's logical to adjust the first image save timer to complete when the image save operation returns.

pkg/image/docker/daemon_provider.go
jonasagx: I thought it would be useful to have that error as a var, but no one references that, I will drop it

PR #102, opened 2022-02-08
Add platform selection

### LINKED ISSUE #101: It should be possible to specify image platform when pulling from the registry

A simple way to specify what platform to pull when a multi-arch image is used.

**Why is this needed**:

A probably a field in `RegistryOptions`.

errordeveloper:
- updata containerd and use it to parse platform string
- refactor `prepareRemoteOptions` a little

Fixes #101

### DIFF
--- a/client.go
+++ b/client.go
@@ -3,7 +3,6 @@ package stereoscope
 import (
 	"context"
 	"fmt"
-
 	"github.com/anchore/stereoscope/internal/bus"
 	dockerClient "github.com/anchore/stereoscope/internal/docker"
 	"github.com/anchore/stereoscope/internal/log"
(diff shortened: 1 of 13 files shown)


wagoodman:
I've updated the branch with a few changes:
- rebased to use functional options for specifying platform (`stereoscope.WithPlatform(string)`)
- Factored out the containerd platform logic without any host defaults. Since stereoscope is on the client side of these requests, using host values as the default os/arch could lead to unintentional edgecases. (Happy to put in an upstream PR, however, it seems that containerd would always want the host defaults, so it didn't seem useful to)
- Expanded platform options to include the `docker.DaemonProvider` (which we use for podman support as well)
- Was hitting the limit on linter method sizes, so some decomposition was done to make the linter happy.

errordeveloper:
@wagoodman thanks a lot for taking this forward, please let me know if there is anything I can help with!

wagoodman:
@errordeveloper absolutely; it would be good to get a second pair of eyes on the platform parsing code I extracted from containerd. That is, I think I have 1:1 logic from what was there with exception made for replacing the default cases for empty strings. 

Also, I'll flip it around: do you see anything that might have been missed? or a missed opportunity while doing this work? 

Lastly, I have a local syft branch that introduces a new `--platform` switch that hasn't been pushed yet, but leverages this. I'll ping you when that's up 👍 

Thanks again for the PR to get this going 🙌 !

spiffcs:
@wagoodman platform code LGTM. 

The other changes don't raise any red flags on my end.

spiffcs (approved):

errordeveloper:
@spiffcs you beat me to it! thanks both for getting this one merged!

PR #104, opened 2022-02-15
Ensure empty files do not have a mimetype

wagoodman:
Today it seems that non-nil readers to files with no contents are resulting in `text/plain` MIME types, which is incorrect. We should return no MIME type. This PR makes this adjustment by ensuring the size of read bytes is not 0.

Related to https://github.com/anchore/syft/pull/782

### DIFF
--- a/pkg/file/mime_type_test.go
+++ b/pkg/file/mime_type_test.go
@@ -2,41 +2,50 @@ package file
 
 import (
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"io"
 	"os"
+	"strings"
 	"testing"
 )
 
 func Test_MIMEType(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/file/mime_type_test.go
```diff
@@ -2,41 +2,50 @@ package file
...
-			// script
-			fixture:  "test-fixtures/mime/capture.sh",
+			name:     "script",
+			fixture:  fileReader("test-fixtures/mime/capture.sh"),
 			expected: "text/plain",
 		},
 		{
-			// no contents
-			fixture:  "",
+			name:     "no contents",
```
jonasagx, line 36: Maybe add a directory as well?

jonasagx (approved):
LGTM with a test suggestion

PR #111, opened 2022-03-12
expand lazy reader interface

jonasagx:
TLDR; Expands lazy reader interface to support io.ReadAt, which is necessary for cataloging golang binaries.

If that interface makes sense for other use cases let's expand it for more catalogers and file resolvers.
Helps with https://github.com/anchore/syft/issues/718
Signed-off-by: Jonas Galvão Xavier <jonas.agx@gmail.com>

### DIFF
--- a/pkg/file/lazy_bounded_read_closer.go
+++ b/pkg/file/lazy_bounded_read_closer.go
@@ -8,7 +8,7 @@ import (
 
 var _ io.ReadCloser = (*lazyBoundedReadCloser)(nil)
 var _ io.ReaderAt = (*lazyBoundedReadCloser)(nil)
-var _ LazyReader = (*lazyBoundedReadCloser)(nil)
+var _ io.Seeker = (*lazyBoundedReadCloser)(nil)
 
 // lazyBoundedReadCloser is a "lazy" read closer, allocating a file descriptor for the given path only upon the first Read() call.
 // Only part of the file is allowed to be read, starting at a given position.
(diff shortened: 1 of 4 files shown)


pkg/file/lazy_read_closer.go
```diff
@@ -7,6 +7,16 @@ import (
 )
 
 var _ io.ReadCloser = (*LazyReadCloser)(nil)
+var _ LazyReader = (*LazyReadCloser)(nil)
+
+// LazyReader is a single interface with all reading functions we want to use
+// across clients. The transition from io.ReadCloser should happen gradually.
+type LazyReader interface {
```
wagoodman, line 14: why is there a new exported interface needed?

jonasagx: I tried it without a new interface, but then the golang cataloger had to cast the reader for each different function call. 
With an exported interface it casts only once -- this interface covers all internal needs.

pkg/file/lazy_bounded_read_closer.go
```diff
@@ -72,3 +64,42 @@ func (d *lazyBoundedReadCloser) Close() error {
...
+func (d *lazyBoundedReadCloser) Seek(offset int64, whence int) (int64, error) {
+	if err := d.isFileOpen(); err != nil {
+		return 0, err
+	}
+
+	return d.reader.Seek(offset, whence)
+}
+
+func (d *lazyBoundedReadCloser) ReadAt(b []byte, off int64) (n int, err error) {
+	if err := d.isFileOpen(); err != nil {
```
wagoodman, line 77: nit: since `isFileOpen` doesn't return a boolean and has side effects a better name might be `openFile`

wagoodman (approved):

PR #120, opened 2022-03-30
Correct docker auth config usage

wagoodman:
This PR makes a few alterations:
- Passes the entire auth config along into docker pull options (not just the username/password portion)
- Some user auth configurations can lead to error messages indicating bad username/password or bad personal access token, even when pulling public images from dockerhub. The fix was to ensure that auth URLs used for `index.docker.io` should resolve to `index.docker.io/v1/`.

Huge thanks to @thaJeztah , @crazy-max, and @chris-crone on troubleshooting this and finding a fix!

### DIFF
--- a/pkg/image/docker/daemon_provider.go
+++ b/pkg/image/docker/daemon_provider.go
@@ -154,33 +154,41 @@ func (p *DaemonImageProvider) pullOptions() (types.ImagePullOptions, error) {
 	}
 	log.Debugf("using docker config=%q", cfg.Filename)
 
-	ref, err := name.ParseReference(p.imageStr)
+	url, err := authURL(p.imageStr)
 	if err != nil {
-		return options, err
+		return options, fmt.Errorf("unable to determine auth hostname: %w", err)
 	}
 
-	hostname := ref.Context().RegistryStr()
-	if hostname == "index.docker.io" {
-		// why do this? There is an upstream issue here: https://github.com/docker/docker-credential-helpers/blob/e595cd69465c6b0f7af2d49582b82fdeddecbf75/wincred/wincred_windows.go#L113-L127
-		// where the hostname used for the auth config lookup requires this or else even pulling public images
-		// will fail with auth related problems (bad username/password, bad personal access token, etc).
-		hostname += "/v1/"
-	}
-
-	authConfig, err := cfg.GetAuthConfig(hostname)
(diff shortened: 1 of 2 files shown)


pkg/image/docker/daemon_provider.go
```diff
@@ -153,30 +154,43 @@ func (p *DaemonImageProvider) pullOptions() (types.ImagePullOptions, error) {
 	}
 	log.Debugf("using docker config=%q", cfg.Filename)
 
-	ref, err := name.ParseReference(p.imageStr)
+	url, err := authURL(p.imageStr)
 	if err != nil {
-		return options, err
+		return options, fmt.Errorf("unable to determine auth hostname: %w", err)
```
kzantow, line 159: nit: is hostname the only possible error here?

pkg/image/docker/daemon_provider.go
```diff
@@ -153,30 +154,43 @@ func (p *DaemonImageProvider) pullOptions() (types.ImagePullOptions, error) {
...
 }
 
+func authURL(imageStr string) (string, error) {
+	ref, err := name.ParseReference(imageStr)
+	if err != nil {
+		return "", err
+	}
+
+	url := ref.Context().RegistryStr()
+	if url == "index.docker.io" {
```
kzantow, line 185: This seems like it might be brittle -- what about self-hosted repos?

pkg/image/docker/daemon_provider_test.go
```diff
@@ -12,11 +17,56 @@ func TestEncodeCredentials(t *testing.T) {
...
+}
 
-	assert.Equal(t, expected, actual, "unexpected output")
+func Test_authURL(t *testing.T) {
+	tests := []struct {
+		imageStr string
+		want     string
+		wantErr  require.ErrorAssertionFunc
+	}{
+		{
```
kzantow, line 45: nit: maybe add a test with a port? `host.io:5000/image:latest`

pkg/image/docker/daemon_provider.go
wagoodman: I should change this reference to "url", but no, they can give a bad image reference as well. I'll change this to a warn instead of an error.

pkg/image/docker/daemon_provider_test.go
wagoodman: done deal

Issue #119, opened 2022-03-30
Wrong source when ":" character in path


jonathongardner:
**What happened**:
When using `image.DetectSource(path)` with `:` in docker archive path `UnknownSource` is returned

**What you expected to happen**:
for `DockerTarballSource` to be returned (or just whatever source the file is if it exists)

**How to reproduce it (as minimally and precisely as possible)**:
```go
package main

import (
  "fmt"

  "github.com/anchore/stereoscope/pkg/image"
)

// run:
// "docker save alpine:3.14.1 > alpine:3.14.1.tar"
// "docker save alpine:3.14.1 > alpine-3.14.1.tar"

func main() {
  s1, l1, _ := image.DetectSource("alpine:3.14.1.tar")
  fmt.Printf("res1: %v-%v\n", s1, l1)

  s2, l2, _ := image.DetectSource("alpine-3.14.1.tar")
  fmt.Printf("res1: %v-%v\n", s2, l2)
}
```

**Anything else we need to know?**:
  Relates to https://github.com/anchore/syft/issues/927

PR #121, opened 2022-03-30
Allow scheme separator in path for detector

jonathongardner:
This is to fix this https://github.com/anchore/stereoscope/issues/119. I ran into this issue when including ":" in a docker archive path and not specifying the scheme. This changes to call `detectSourceFromPath` if the scheme hint is invalid. I think this should be safe (for something like a docker image reference) since `detectSourceFromPath` method checks to see if the path exists.


jonasagx:
Hi @jonathongardner thank you for reporting the bug and a solution! Your change looks good, I tested it locally and it works. Before approving it, I would like to run our CI validations which are not working for this PR. 
I will try closing and reopening this PR to trigger the validations. I apologize for extra friction.

jonasagx (approved):
Thank you for your contribution @jonathongardner 🎉

PR #122, opened 2022-04-06
Retry auth URL lookup without docker credentialhelper workaround

wagoodman:
Addresses https://github.com/anchore/syft/issues/936

This PR builds on https://github.com/anchore/stereoscope/pull/120, where we tailor URLs for dockerhub when authenticating against the registry (via the docker daemon), however, this original change was meant to fix behavior when using the credential helpers. This change regressed behavior for users that have `auth` entries directly in the `~/.docker/config.json`. This PR changes the auth behavior to search for credentials with the credentials workaround first, and if not found, then fallback behavior is to try the credential search again with the original url.

### DIFF
--- a/pkg/image/docker/daemon_provider.go
+++ b/pkg/image/docker/daemon_provider.go
@@ -154,7 +154,8 @@ func (p *DaemonImageProvider) pullOptions() (types.ImagePullOptions, error) {
 	}
 	log.Debugf("using docker config=%q", cfg.Filename)
 
-	url, err := authURL(p.imageStr)
+	// get a URL that works with docker credential helpers
+	url, err := authURL(p.imageStr, true)
 	if err != nil {
 		log.Warnf("failed to determine auth url from image=%q: %+v", p.imageStr, err)
 		return options, nil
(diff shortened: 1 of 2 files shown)


spiffcs (approved):

pkg/image/docker/daemon_provider.go
```diff
@@ -176,17 +196,19 @@ func (p *DaemonImageProvider) pullOptions() (types.ImagePullOptions, error) {
...
+func authURL(imageStr string, dockerhubWorkaround bool) (string, error) {
 	ref, err := name.ParseReference(imageStr)
 	if err != nil {
 		return "", err
 	}
 
 	url := ref.Context().RegistryStr()
-	if url == "index.docker.io" {
+	if dockerhubWorkaround && url == "index.docker.io" {
 		// why do this? There is an upstream issue here: https://github.com/docker/docker-credential-helpers/blob/e595cd69465c6b0f7af2d49582b82fdeddecbf75/wincred/wincred_windows.go#L113-L127
```
spiffcs, line 207: For people coming to this PR. This comment gives a lot of the context as to why this whole daemon provider works the way it does.

PR #123, opened 2022-04-06
Add client method to support sylabs SIF images

Poluect:
This PR adds a separate method, `GetSifImageFromSource`, to support sylabs SIF images (currently only files from disk). `sif-file` source needs to be explicitly specified in user input

Main challenge of adding SIF support to existing `GetImageFromSource` is the difference between oci, docker images file scanning mechanism and SIF. In current implementation SIF file is mounted to the temporary directory using `squashfuse`, so usual directory scanning needs to be applied for further creating of an SBOM, which differs from oci and docker scanning of layers using `go-containerregistry` package structs. 

`GetSifImageFromSource` returns `sif.Image` struct, which contains image metadata and path of a mounted directory, which then can by used by `syft` for creating of an SBOM

Appreciate any feedback

Related: https://github.com/anchore/syft/issues/937


wagoodman:
This is really interesting, thanks for submitting! My biggest hesitation about the implementation is the need to mount and unmount the image. We've strived to keep stereoscope as close to static analysis as possible, so for OCI images and understanding the squash representation of an image, we don't create containers or any overlayfs mounts, but instead build [a filetree based off of the underlying tars](https://github.com/anchore/stereoscope/blob/main/pkg/image/layer.go#L144-L198). Is it possible to do the same thing with a sif image? Happy to work with you on this!

wagoodman:
I'm going to close this PR in favor of https://github.com/anchore/stereoscope/pull/125 ... if I'm mistaken shout out and I can reopen for discussion.

PR #124, opened 2022-04-10
Add locking to FileCatalog

ajvpot:
Adds a sync.RWMutex to FileCatalog to prevent concurrent use of `source.FileResolver` on the same image from causing a concurrent map write panic.

This is a prerequisite to making Catalogers safe to use concurrently.


wagoodman (approved):
Thanks for the improvement!

PR #125, opened 2022-04-27
Singularity Image Support

tri-adam:
This PR is based on the work in #123, and is proposed as a potential way forward based on the feedback there and the discussion in the last community call. @Poluect has unfortunately had to step away from the Sylabs team on short notice, so I'm picking up where he left off.

This PR uses [github.com/CalebQ42/squashfs](https://pkg.go.dev/github.com/CalebQ42/squashfs) to read the root filesystem contents. This is a pure Go library, and doesn't require a `mount`.

Structurally, this PR takes a different approach than #123, exposing Singularity Images via the existing `func GetImageFromSource` rather than adding a SIF-specific `func GetSifImageFromSource`. This adds a bit more complexity in `stereoscope`, but keeps the exported API tighter and makes it a lot easier to support SIF in projects consuming this module such as `syft` (I'll open a PR there shortly that uses pulls in this work.)

Related to https://github.com/anchore/syft/issues/937, https://github.com/sylabs/sif/issues/190


tri-adam:
Hi @wagoodman, just checking in to see if there's anything I can do to push this forward. Thanks!

jonasagx:
@tri-adam there are issues with the licenses from these dependencies:
```
$ make static-analysis
Allow Rules: [BSD.* MIT.* Apache.* MPL.*]
Unallowable license (GPL-2.0) from "github.com/rasky/go-lzo"
Unallowable license (CC0-1.0) from "github.com/therootcompany/xz"
```

wagoodman:
@tri-adam we're going to review these new licenses internally to ensure we can pull in these dependencies safely (from a legal perspective), I'll report back soon

tri-adam:
> Hey @tri-adam thank you for working on this feature, I ran CI validations and there are issues with licenses from these dependencies:
> 
> ```
> $ make static-analysis
> Allow Rules: [BSD.* MIT.* Apache.* MPL.*]
> Unallowable license (GPL-2.0) from "github.com/rasky/go-lzo"
> Unallowable license (CC0-1.0) from "github.com/therootcompany/xz"
> ```

> @tri-adam we're going to review these new licenses internally to ensure we can pull in these dependencies safely (from a legal perspective), I'll report back soon

@wagoodman thanks. It looks like both of those are pulled in by github.com/CalebQ42/squashfs:

```
$ go mod why github.com/rasky/go-lzo github.com/therootcompany/xz
# github.com/rasky/go-lzo
github.com/anchore/stereoscope/pkg/file
github.com/CalebQ42/squashfs
github.com/CalebQ42/squashfs/internal/compression
github.com/rasky/go-lzo

# github.com/therootcompany/xz
github.com/anchore/stereoscope/pkg/file
github.com/CalebQ42/squashfs
github.com/CalebQ42/squashfs/internal/compression
github.com/therootcompany/xz
```

Let me know the results of the legal review, thanks.

wagoodman:
Hey @tri-adam , sorry to report that GPL-2.0 licensed dependencies cannot be introduced into the codebase at this time (CC0-1.0 is ok to introduce though). Is seems that the LZO license is not something that could be controlled by the repo owner too:

> Being a straightforward port of the original source code, it shares the same license (GPLv2) as I can't possibly claim any copyright on it.

There are a couple of paths forward:
1. we can try to find an lzo alternative go implementation (I couldn't find one after a quick search)
2. roll our own implementation of LZO (this looks pretty hard, given that LZO is a collection of a dozen or so algorithms that are very poorly documented)
3. is LZO required for the SIF image functionality? if not, then we could fork the squashfs lib and remove the dependency? (haven't dug into this option at all to check feasibility)
4. we could head back down the `mount` PR and work more closely on an opt-in approach
5. _<insert-other-idea-here>_ <-- open to more suggestions!

Thanks for bearing with us on this 🙏

tri-adam:
> There are a couple of paths forward:
> 
> 1. we can try to find an lzo alternative go implementation (I couldn't find one after a quick search)
> 2. roll our own implementation of LZO (this looks pretty hard, given that LZO is a collection of a dozen or so algorithms that are very poorly documented)
> 3. is LZO required for the SIF image functionality? if not, then we could fork the squashfs lib and remove the dependency? (haven't dug into this option at all to check feasibility)
> 4. we could head back down the `mount` PR and work more closely on an opt-in approach
> 5. _insert-other-idea-here_ <-- open to more suggestions!
> 
> Thanks for bearing with us on this 🙏

Took a stab at option 3, removing LZO support from the underlying module, and it's reasonably easy (https://github.com/CalebQ42/squashfs/compare/v0.5.4...sylabs:remove-lzo-v0.5.4).

SingularityCE does not appear to rely on LZO compression, so I don't foresee any negative impacts in terms of functionality.

Still some further testing to do, but it's a start!

wagoodman:
> SingularityCE does not appear to rely on LZO compression, so I don't foresee any negative impacts in terms of functionality.

great news!

go.mod
```diff
@@ -20,11 +21,15 @@ require (
...
 	github.com/stretchr/testify v1.7.0
+	github.com/sylabs/sif/v2 v2.7.0
 	github.com/wagoodman/go-partybus v0.0.0-20200526224238-eb215533f07d
 	github.com/wagoodman/go-progress v0.0.0-20200621122631-1a2120f0695a
-	golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2
+	golang.org/x/crypto v0.0.0-20210921155107-089bfa567519
 )
+
+// Forked to remove https://github.com/rasky/go-lzo dependency, which is GPLv2 licensed.
+replace github.com/CalebQ42/squashfs => github.com/sylabs/squashfs v0.5.5-0.20220526223455-67e0f4cd95c5
```
wagoodman, line 35: we tend to not use `replace` directives since it makes build reproducibility harder -- mind swapping out the `require` statement

pkg/image/source.go
```diff
@@ -76,6 +79,8 @@ func ParseSourceScheme(source string) Source {
 		return OciTarballSource
 	case "oci-registry", "registry":
 		return OciRegistrySource
+	case "singularity-image":
```
wagoodman, line 82: I think we could use a shorter name here such as `singularity` or `sif` (and "image" is redundant here) -- what do you think @tri-adam ?

go.mod
wagoodman: I take back my comment -- I see the fork is keeping the same module name

wagoodman (commented):
FYI -- force pushed to resolve merge conflict in go.mod/go.sum

wagoodman:
@tri-adam right now integration tests are failing due to lack of image types covered (since singularity was added). The next step is to add testing that covers:
- symlinks: https://github.com/anchore/stereoscope/blob/9e3b40f4a9d3a75b25868b2d1ef0fdcb8604e212/test/integration/fixture_image_symlinks_test.go#L30-L53
- a parsing go case: https://github.com/anchore/stereoscope/blob/9e3b40f4a9d3a75b25868b2d1ef0fdcb8604e212/test/integration/fixture_image_simple_test.go#L25-L61

I think the one thing I'm hung up on is that these tests depend on building the same image and obtaining it in different ways -- is there a way to nicely achieve that in testing with singularity images? If not, that's ok, we can make one off tests to assert this functionality (in integration or unit-level testing).

pkg/image/source.go
dtrudg: `sif` would likely be my suggestion - as that's what has been adopted by support brought into `containers/image` and hence `podman` which are using `sif` as the 'transport' name.

dtrudg:
Hi @wagoodman. Thanks for the review and comments on this! I work with @tri-adam at Sylabs and might be picking this work up from him for a short time.

I'll look into the tests in detail as soon as possible. With regard to building a SIF, within the workflow, that would be possible from the docker image that is generated here with an install of Singularity. We do package SingularityCE ourselves for Ubuntu, with those packages hosted on GitHub releases ... but I'm not sure whether you'd want to add an install into the CI workflows?

A final thought is that we could get close enough to a 'real' Singularity built SIF by directly leveraging `mksquashfs` and calling the `github.com/sylabs/sif` SIF code we are importing to construct a simple SIF.... but perhaps this is a bit too far removed from the intent to test against a real Singularity container image?

wagoodman:
hey @dtrudg 👋 

> a. [it] would be possible from the docker image that is generated here with an install of Singularity...I'm not sure whether you'd want to add an install into the CI workflows?

> b. ... we could get close enough to a 'real' Singularity built SIF by directly leveraging `mksquashfs` and calling the github.com/sylabs/sif SIF code we are importing to construct a simple SIF.

I'm ok with either approach, my only hesitation with option `a` is if the install time is long enough to make PR validations much slower --but if that is the more kosher approach then that might be an OK trade off. 

The goal of any approach is to make certain that the content API of stereoscope functions with the new image (querying files by a glob or path, getting file contents, resolving symlinks properly, etc). 

What's the approach that you suggest / lean towards?

dtrudg:
I'm sorry for the test failure due to the apt install... that was a :facepalm:

dtrudg:
Hi - I've re-based again onto the latest main. Looks like the last CI run prior to the rebase did pass, so would be grateful for any feedback on the testing approach, and what cleanup might be needed here. Many thanks!

wagoodman:
sorry for the delay @dtrudg , approved the workflow and are reviewing the changes now.

wagoodman (approved):
The only change I made was making the singularity call more portable to help development on macs -- nothing too major. Thanks for the enhancement and updating the tests @tri-adam @dtrudg !!

dtrudg:
Many thanks @wagoodman - I'll take a look at the related syft PR next.

Thanks for your help, and apologies this was pretty slow form our side.

wagoodman:
@dtrudg no need to apologize, this slid on my side as well 👍 I'm looking at https://github.com/anchore/syft/pull/974 and just pulled in the latest stereoscope and I think I'm seeing some integration issues, but too soon to tell really. I'll dig further and see what comes up.

wagoodman:
@dtrudg I updated https://github.com/anchore/syft/pull/974 and it seems to be working great -- did you want to add anything or give it a shot before merging?

dtrudg:
> @dtrudg I updated [anchore/syft#974](https://github.com/anchore/syft/pull/974) and it seems to be working great -- did you want to add anything or give it a shot before merging?

@wagoodman I'll take a quick check within the next couple of hours when I'm out of a couple of calls, and ping you over on that PR if that's okay? I think it should be good but don't want to overlook anything given I'm picking it up from tri-adam :-)

PR #134, opened 2022-08-03
Remove replace directive for sylabs/squashfs

kzantow:
This PR updates the reference to `sylabs` fork of `squashfs` and removes the `replace` directive to use it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/anchore/stereoscope
 go 1.16
 
 require (
-	github.com/CalebQ42/squashfs v0.5.4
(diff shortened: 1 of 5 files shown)


dtrudg:
Hi @kzantow - apologies I didn't see this before I opened #135 

It looks like here that you updated to the head of the `remove-lzo` branch on sylabs/squashfs. This is the upsteam main, with lzo support removed. However the upstream main seems to have breaking changes. hence in #135 I'm using the head of the `remove-lzo-v0.5.4` branch, which I've also tagged as `v0.5.4+remove.lzo.1` now.

kzantow:
@dtrudg aha! thanks much! we can definitely go with your PR

PR #136, opened 2022-08-04
add latest as default tag for container images

Ferada:
Fix for <https://github.com/anchore/syft/issues/411>, according to [this comment](https://github.com/anchore/syft/issues/411#issuecomment-897676650), not sure that's the right spot though.

Also, this method will then resolve the container name to the fully expanded version, including the registry, I'm not sure that doesn't interfere with anything, just to mention it.

The OCI version already works, no change needed there.

### DIFF
--- a/client.go
+++ b/client.go
@@ -114,13 +114,19 @@ func selectImageProvider(imgStr string, source image.Source, cfg config) (image.
 		if err != nil {
 			return nil, err
 		}
-		provider = docker.NewProviderFromDaemon(imgStr, tempDirGenerator, c, cfg.Platform)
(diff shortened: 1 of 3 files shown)


spiffcs:
Thanks for the PR @Ferada!

wagoodman (approved):
Thanks! I added a few tests and changed the behavior some to not assume a registry -- we try not to make any assumptions based on the input and allow the tooling (in this case docker) to resolve any defaults here (notable exception is the tag of course).

Ferada:
> I added a few tests and changed the behavior some to not assume a registry -- we try not to make any assumptions based on the input and allow the tooling (in this case docker) to resolve any defaults here (notable exception is the tag of course).

Makes sense, thanks.

PR #142, opened 2022-09-01
Auto-detect Singularity images

tri-adam:
Modify `detectSourceFromPath` to auto-detect Singularity container images. Re-factor `pkg/image` unit tests to accommodate more thorough SIF test cases, and extend unit tests to cover new functionality.

I believe this may address https://github.com/anchore/syft/issues/1189.


pkg/image/source_test.go
```diff
@@ -10,50 +10,56 @@ import (
...
+			name:  "docker-engine-impossible-id",
+			getFS: getDummyEmpty(),
 			// not a valid ID
 			input:            "a5E",
 			source:           UnknownSource,
 			expectedLocation: "",
 		},
 		{
 			name:             "docker-engine",
+			getFS:            getDummyEmpty(),
```
kzantow, line 55: maybe just make this the default instead of needing to add it to every test case?

kzantow (commented):
Just a couple more comments here -- hopefully which make sense! I'm thinking about maintenance and having a pattern of returning callbacks makes things a bit harder to debug from my experience.

pkg/image/source_test.go
```diff
@@ -282,82 +292,81 @@ func TestParseScheme(t *testing.T) {
...
 		expectedSource Source
-		sourceType     string
 		expectedErr    bool
 	}{
 		{
 			name:           "no tar paths",
-			paths:          []string{},
-			sourceType:     "tar",
+			path:           "image.tar",
+			getFS:          getDummyTar("image.tar"),
```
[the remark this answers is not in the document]
tri-adam: I'm not sure it can be avoided for all test cases. For example, in the `no path given` test case, `path` is set to `"/does-not-exist"` even though the `afero.Fs` is empty.

pkg/image/source_test.go
```diff
@@ -282,82 +292,81 @@ func TestParseScheme(t *testing.T) {
...
+			name:           "singularity-path",
+			path:           "image.sif",
+			getFS:          getDummySIF("image.sif"),
+			expectedSource: SingularitySource,
+		},
 	}
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			fs := afero.NewMemMapFs()
```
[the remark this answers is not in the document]
tri-adam: Sure, yeah will do!

pkg/image/source_test.go
```diff
@@ -104,57 +104,67 @@ func TestDetectSource(t *testing.T) {
...
+			source:           SingularitySource,
+			expectedLocation: "~/a-potential/path.sif",
+		},
+		{
+			name:             "singularity-path-explicit",
+			fs:               getDummySIF(t, "~/a-potential/path.sif"),
+			input:            "singularity:~/a-potential/path.sif",
+			source:           SingularitySource,
+			expectedLocation: "~/a-potential/path.sif",
 		},
```
kzantow, line 161: Is this the exact same test as `singularity-path` above?

kzantow (approved):
LGTM -- just one more minor comment

pkg/image/source_test.go
tri-adam: They are similar, but they test different functions (`detectSource` vs `detectSourceFromPath`.) `detectSource` calls `detectSourceFromPath` in some instances (when the source scheme is omitted), but not in others. For that reason, I think it's _probably_ best to test both explicitly, as is done with the other source types?

kzantow:
Thanks for the contribution, @tri-adam!

PR #130, opened 2022-06-16
rollback containerd to v1.5.13

spiffcs:
In order to release syft and grype we need a stable dependency graph.

Currently, the bump of containerd v1.6.6 pulls in a version of opentelemetry(v1.7.0) that is incompatible with underlying libraries

While we work to fix this conflict among the different libraries we depend on this rollback will unblock the release and allow syft/grype to update to a version of stereoscope not affected by [CVE-2022-31030](https://github.com/advisories/GHSA-5ffw-gxpp-mxpf)

Signed-off-by: Christopher Phillips <christopher.phillips@anchore.com>

PR #144, opened 2022-10-03
bump containerd to v1.6.8

westonsteimel:
Awhile back we tried bumping containerd to the v1.6.x series in sterescope, but ended up having to roll it back to v1.5.x in https://github.com/anchore/stereoscope/pull/130 due to dependency resolution conflicts with other libraries when trying to update stereoscope in syft.  This is to check if the dependencies cleanly resolve now on the latest v1.6.8 release of containerd.


westonsteimel:
Hooray, looks like dependency resolution is clean in both syft and grype when doing this update to containerd v1.6.8!
- https://github.com/anchore/syft/pull/1237
- https://github.com/anchore/grype/pull/943
The [unit test failure in grype](https://github.com/anchore/grype/actions/runs/3172668284/jobs/5167391915#step:8:58) appears unrelated to the containerd bump and seems to be around the ConanMetadata, which may just be an artifact of bringing in currently unreleased syft changes in order to test.

spiffcs (approved):

PR #147, opened 2022-11-30
Upgrade logger to anchore/go-logger interface

wagoodman:
Swaps the current logger interface for the [anchore/go-logger](github.com/anchore/go-logger) interface


client.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/anchore/go-logger"
```
kzantow, line 7: nit: it would be nice if we were using `gosimports` here

kzantow (approved):

client.go
wagoodman: I agree, I can follow up in another PR to get stereoscope up to date with some of the other repos

PR #151, opened 2023-01-11
Add new Path filter function to index paths based on the filter

gnmahanth:
Fix: https://github.com/anchore/syft/issues/1446

### DIFF
--- a/client.go
+++ b/client.go
@@ -67,7 +67,7 @@ func WithPlatform(platform string) Option {
 }
 
 // GetImageFromSource returns an image from the explicitly provided source.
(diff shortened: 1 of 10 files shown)


wagoodman (commented):
I like the improvement to the visitor 👍 I have a couple of comments:
1. My preference on the changes at the top-level API is to lean into the existing functional options (type `Option`) approach and allow for a filter to be optional. In that sense, passing a default function that always returns true is alright, but a slightly more preferable approach would be to pass `nil` by default and not call the filter if it doesn't exist down in the visitor.
2. A test is needed to show the effect of the path filter on the visitor.

Shout out if you want to chat more about this!

wagoodman:
@gnmahanth would you like help incorporating the above comments?

PR #152, opened 2023-01-12
Set the default platform for select sources based on host arch

### LINKED ISSUE #149: Stereoscope pulls different images when using ``docker:`` vs ``registry:`` from multi-platform images
**What happened**:
On an arm machine, stereoscope produces different results when pulling a multi-platform image using ``docker`` scheme vs ``registry`` scheme.
The ``docker`` scheme correctly infers the host architecture and pulls the correct matching image whereas the ``registry`` scheme pulls only ``amd64/linux`` image when platform is not explicitly provided.

The platform for both images should be the same, preferably the host platform.
If inference of platform is not intended in an OCI/registry source, this difference should be verbose in debug logs or in documentation.

**How to reproduce it (as minimally and precisely as possible)**:
On an arm machine, ``registry`` scheme yields an ``x86`` image (incorrectly)
```
; go run examples/basic.go registry:rust:1.42-slim-buster | grep binutils-x86
[0000] DEBUG image: source=OciRegistry location=rust:1.42-slim-buster
[0000] DEBUG pulling image info directly from registry image="rust:1.42-slim-buster"
[0000] DEBUG no registry credentials configured, using the default keychain
[0003] DEBUG image metadata: digest=sha256:165a37f9b454c1b000732bc6fb31edb6710818f6096796e7951a1e91cafa758d mediaType=application/vnd.docker.distribution.manifest.v2+json tags=[]
[0003] DEBUG layer metadata: index=0 digest=sha256:b60e5c3bcef2f42ec42648b3acf7baf6de1fa780ca16d9180f3b4a3f266fe7bc mediaType=application/vnd.docker.image.rootfs.diff.tar.gzip
[0005] DEBUG layer metadata: index=1 digest=sha256:508d8706fc1db20449e5e0fe22c65e1cce8ee30d062c5357f7e81f22cb8c3441 mediaType=application/vnd.docker.image.rootfs.diff.tar.gzip
 ...
    /usr/share/doc/binutils-x86-64-linux-gnu
 ...
```

On the same machine, ``docker`` scheme yields an ``arm`` image (correctly)
```
; go run examples/basic.go docker:rust:1.42-slim-buster | grep binutils-aarch64
[0000] DEBUG image: source=DockerDaemon location=rust:1.42-slim-buster
[0008] DEBUG image metadata: digest=sha256:8dc750c34fb3e46a4585e531416d995592fd7ca89630be76cca23ce1da26391f mediaType=application/vnd.docker.distribution.manifest.v2+json tags=[rust:1.42-slim-buster]
[0008] DEBUG layer metadata: index=0 digest=sha256:67d3a85d42a2b4ae3ca54ff7b6225bb136fbe0fa5732d9c5143470f13470bbde mediaType=application/vnd.docker.image.rootfs.diff.tar.gzip
[0008] DEBUG layer metadata: index=1 digest=sha256:081726014efb2e4d440405aee79d2cca3966c3fc965ce2cee13fd972c907e953 mediaType=application/vnd.docker.image.rootfs.diff.tar.gzip
    /usr/share/doc/binutils-aarch64-linux-gnu
...
```

``DaemonImageProvider`` infers host platform and pulls the correct image whereas
``RegistryImageProvider`` uses [go-containerregistry](https://github.com/google/go-containerregistry) which defaults to ``amd64/linux`` when platform is missing rather than inferring the host platform.
See also:
- https://github.com/google/go-containerregistry/issues/1505

**Environment**:
- OS: MacOS Monterey, M1 Pro.

wagoodman:
This PR sets a default platform to image providers that support platform options only if there is no overriding platform specified by the caller.

TODO:
- [ ] add tests (which is why this is in draft)

Fixes #149


client.go
```diff
@@ -129,28 +130,45 @@ func selectImageProvider(imgStr string, source image.Source, cfg config) (image.
...
+	defaultPlatform, err := image.NewPlatform(defaultArch)
+	if err != nil {
+		log.WithFields("error", err).Warnf("unable to set default platform to %q", runtime.GOARCH)
+		defaultPlatform = nil
+	}
+
+	switch source {
+	case image.DockerTarballSource, image.OciDirectorySource, image.OciTarballSource, image.SingularitySource:
+		if cfg.Platform != nil {
+			return fmt.Errorf("specified platform=%q however image source=%q does not support selecting platform", cfg.Platform.String(), source.String())
```
kzantow, line 158: Does this need to be an error? could it just safely be ignored here?

client.go
```diff
@@ -129,28 +130,45 @@ func selectImageProvider(imgStr string, source image.Source, cfg config) (image.
...
-			return nil, platformSelectionUnsupported
-		}
 		provider = sif.NewProviderFromPath(imgStr, tempDirGenerator)
 	default:
 		return nil, fmt.Errorf("unable to determine image source")
 	}
 	return provider, nil
 }
 
+func setPlatform(source image.Source, cfg *config, defaultArch string) error {
```
kzantow, line 146: nit: `setDefaultPlatform` or similar, since it does nothing if platform is set?

kzantow (approved):

client.go
wagoodman: Well, on second thought, this is also verifying that a non-default, non-nil platform is allowed to be set by the user... which goes beyond setting the default. There is probably a better name for this but setPlatform and setDefaultPlatform both don't quite fit the bill... but setPlatform is slightly less incorrect, so I'll leave it for now.

PR #153, opened 2023-01-18
Evaluation of memoziation for p.Normalize()

aeg:
This PR shows evaluation code for memoziation for p.Normalize()

It can improve the performance of syft.

Issue: https://github.com/anchore/syft/issues/1480


wagoodman:
@aeg would you like help incorporating the above comments?

aeg:
> would you like help incorporating the above comments?

sorry, I think this pull request is out of date.
As a result of the syft version increase, p.normalize() does not seem to be a bottleneck in the current (0.85.0).

I ran the benchmark again and it wasn't even in the top 20.

I want close this pr.

Thanks,

PR #154, opened 2023-01-24
Add additional catalog indexes for performance

wagoodman:
This adds additional indexes for:
- basename
- basename globs
- file extension

These can be optionally leveraged for improved performance instead of using the standard search by glob against the filetree.

Preview of change in syft (today's behavior):
```
time go run ./cmd/syft ~/images/virt-manager.tar -vvv  
# ...
real	4m34.067s
user	5m49.291s
sys	0m6.189s
```

With the new indexes being leveraged:
```
time go run ./cmd/syft ~/images/virt-manager.tar -vvv
# ...
real	38.155s
user	43.73s
sys	12.88s
```


pkg/image/content_helpers.go
```diff
@@ -31,20 +31,64 @@ func fetchFileContentsByPath(ft *filetree.FileTree, fileCatalog *FileCatalog, pa
 func fetchFilesByMIMEType(ft *filetree.FileTree, fileCatalog *FileCatalog, mType string) ([]file.Reference, error) {
 	fileEntries, err := fileCatalog.GetByMIMEType(mType)
 	if err != nil {
-		return nil, fmt.Errorf("unable to fetch file references by MIME type: %w", err)
+		return nil, fmt.Errorf("unable to fetch file references by MIME type (%q): %w", mType, err)
```
spiffcs, line 34: %q: a single-quoted character literal safely escaped with Go syntax

Nice

pkg/image/content_helpers.go
```diff
@@ -31,20 +31,64 @@ func fetchFileContentsByPath(ft *filetree.FileTree, fileCatalog *FileCatalog, pa
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch file references by basename (%q): %w", basename, err)
+	}
+
+	return filterCatalogFilesRelativesToTree(ft, fileEntries, false, filetree.FollowBasenameLinks)
+}
+
+// fetchFilesByBasenameGlob is a common helper function for resolving file references for a file basename glob pattern
+// catalog relative to the given tree.
+func fetchFilesByBasenameGlob(ft *filetree.FileTree, fileCatalog *FileCatalog, basenameGlob string) ([]file.Reference, error) {
```
spiffcs, line 65: I know we're trying to move indexes away from `Glob` matching in this PR - Is this the best name here?

The function is written so that `**` and `/` are not supported so I'm trying to think through what kinds of globs are matched then which improve the performance. Let me also check out the test for this since that might help in my review. I'l follow up with other comments as I go through this.

pkg/image/file_catalog.go
```diff
@@ -107,3 +209,27 @@ func (c *FileCatalog) FileContents(f file.Reference) (io.ReadCloser, error) {
...
+		return exts
+	}
+
+	// ignore directories
+	if strings.HasSuffix(p, "/") {
+		return exts
+	}
+
+	// ignore . which indicate a hidden file
+	p = strings.TrimLeft(path.Base(p), ".")
```
spiffcs, line 228: The comment says we are ignoring hidden files here, but the later logic still looks like that when `.` is found we append the extension of the hidden file.

spiffcs (commented):
Initial LGTM - I need to go back and double check the test file since it was the largest diff from this change. The tests run correctly locally and no diffs between local env and CI.

Do you want to wait for the syft PR that incorporates these changes before doing the final merge?

I can also test the syft integration if that's up for review

pkg/image/file_catalog.go
wagoodman: the later logic starts from the end of the string and moves to the front of the string, reporting all extensions. So for instance, `file.tar.gz` would result in `.tar.gz` and `.gz` being found in the below loop. This "trim" protection is protecting against input like `.file.tar.gz` or `...file.tar.gz` so that we still only return `.tar.gz` + `gz` (and not something like `.file.tar.gz`)

kzantow (dismissed):

wagoodman:
@spiffcs 

> Do you want to wait for the syft PR that incorporates these changes before doing the final merge?

Already done! https://github.com/anchore/syft/pull/1510 (all PR checks passing, a lot of tests added)

pkg/image/content_helpers.go
```diff
@@ -8,43 +8,20 @@ import (
 	"github.com/anchore/stereoscope/pkg/filetree"
 )
 
-// fetchFileContentsByPath is a common helper function for resolving the file contents for a path from the file
+// fetchReaderByPath is a common helper function for resolving the file contents for a path from the file
 // catalog relative to the given tree.
-func fetchFileContentsByPath(ft *filetree.FileTree, fileCatalog *FileCatalog, path file.Path) (io.ReadCloser, error) {
-	exists, fileReference, err := ft.File(path, filetree.FollowBasenameLinks)
+func fetchReaderByPath(ft filetree.Reader, fileCatalog FileCatalogReader, path file.Path) (io.ReadCloser, error) {
```
kzantow, line 13: I'm 👍 on this naming change here, but it might be easier to review with fewer "cleanup" changes

pkg/file/resolution.go
```diff
@@ -0,0 +1,158 @@
...
+import (
+	"sort"
+
+	"github.com/scylladb/go-set/strset"
+)
+
+// Resolution represents the fetching of a possibly non-existent file via a request path.
+type Resolution struct {
+	RequestPath Path
+	*Reference
```
spiffcs, line 12: Is reference `Reference` what is fetched here? Is this why we use a pointer since it can be non-existent?

pkg/image/image_test.go
```diff
@@ -97,13 +98,17 @@ func TestImageAdditionalMetadata(t *testing.T) {
...
 
-			img := NewImage(nil, tempFile.Name(), test.options...)
+			img := New(nil, tempFile.Name(), test.options...)
 
 			err = img.applyOverrideMetadata()
 			if err != nil {
 				t.Fatalf("could not create image: %+v", err)
 			}
-			for _, d := range deep.Equal(img, &test.image) {
+			if d := cmp.Diff(img, &test.image,
```
kzantow, line 107: nit: could an `assert.JSONEq` work here? The output can be easier to read than the `cmp.Diff`

pkg/filetree/builder.go
```diff
@@ -0,0 +1,54 @@
...
+
+func NewBuilder(tree Writer, index IndexWriter) *Builder {
+	return &Builder{
+		tree:  tree,
+		index: index,
+	}
+}
+
+func (b *Builder) Add(metadata file.Metadata) (*file.Reference, error) {
+	var ref *file.Reference
```
spiffcs, line 23: nit:
```
var
var

vs 

var (

)
```

Up to you =)

pkg/filetree/depth_first_path_walker.go
```diff
@@ -59,11 +59,12 @@ func (w *DepthFirstPathWalker) Walk(from file.Path) (file.Path, *filenode.FileNo
 	w.pathStack.Push(from)
 
 	var currentPath file.Path
-	var currentNode *filenode.FileNode
+	var currentNode *nodeAccess
```
spiffcs, line 62: nit: style guide shows `var ()` for multiple declarations, but we've never really mentioned it before now:
https://google.github.io/styleguide/go/best-practices#variable-declarations

pkg/filetree/depth_first_path_walker.go
```diff
@@ -59,11 +59,12 @@ func (w *DepthFirstPathWalker) Walk(from file.Path) (file.Path, *filenode.FileNo
...
 	var err error
 
 	for w.pathStack.Size() > 0 {
 		currentPath = w.pathStack.Pop()
+		// TODO: should we make these link resolutions configurable so you can observe the links on walk as well?
```
spiffcs, line 67: Is this TODO still a candidate for this PR or is it a future consideration? 

What do you mean by configurable? Is it a setting where instead `true` are replaced by some user provided value?

pkg/image/image.go
```diff
@@ -214,31 +223,38 @@ func (i *Image) Read() error {
 	i.Layers = layers
 
 	// in order to resolve symlinks all squashed trees must be available
-	return i.squash(readProg)
+	err = i.squash(readProg)
```
spiffcs, line 226: Note: I want to come back to this on a second pass:

Should `i.SquashedTreee()` call `i.squash()`?

Does squash() affect `fileCatalog`?

Coming back to look here in a second

pkg/image/layer.go
```diff
@@ -80,7 +81,8 @@ func (l *Layer) uncompressedTarCache(uncompressedLayersCacheDir string) (string,
 // file tree, and the layer squash tree.
 func (l *Layer) Read(catalog *FileCatalog, imgMetadata Metadata, idx int, uncompressedLayersCacheDir string) error {
 	var err error
-	l.Tree = filetree.NewFileTree()
+	tree := filetree.New()
```
spiffcs, line 84: Is there any difference here doing `l.Tree = filetree.New()` and then passing `l.tree` to downstream functions?

I noticed `NewSearchContext()` takes `l.Tree`but things like `squashfsVisitor takes tree`

test/integration/fixture_image_simple_test.go
```diff
@@ -190,7 +190,8 @@ func BenchmarkSimpleImage_FetchSquashedContents(b *testing.B) {
 
 func assertImageSimpleMetadata(t *testing.T, i *image.Image, expectedValues testCase) {
 	t.Helper()
-	t.Log("Asserting metadata...")
+	//t.Log("Asserting metadata...")
```
spiffcs, line 193: nit: this is commented out in the test - do we want to keep this or delete it?

test/integration/fixture_image_simple_test.go
```diff
@@ -226,21 +227,23 @@ func assertImageSimpleMetadata(t *testing.T, i *image.Image, expectedValues test
 }
 
 func assertImageSimpleSquashedTrees(t *testing.T, i *image.Image) {
-	t.Log("Asserting squashed trees...")
-	one := filetree.NewFileTree()
+	t.Helper()
+	//t.Log("Asserting squashed trees...")
```
spiffcs, line 231: nit: more t.Log comment out

test/integration/fixture_image_simple_test.go
```diff
@@ -263,18 +266,20 @@ func assertImageSimpleSquashedTrees(t *testing.T, i *image.Image) {
 }
 
 func assertImageSimpleTrees(t *testing.T, i *image.Image) {
-	t.Log("Asserting trees...")
-	one := filetree.NewFileTree()
+	t.Helper()
+	//t.Log("Asserting trees...")
```
spiffcs, line 270: nit: t.Log 

Reading more of these - are we keeping these for the developer's use case to uncomment debugging test

spiffcs (commented):
First pass done - I've got about 23 files left I've marked to go back through to figure out how they work.

pkg/file/id_set.go
```diff
@@ -0,0 +1,75 @@
+//nolint:dupl
+package file
+
+import "sort"
+
+type IDSet map[ID]struct{}
```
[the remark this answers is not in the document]
wagoodman: I already did, I ran into a 1.19 specific generics bug that causes it to fail under a specific circumstance. For that reason I reverted and went back to plain-ol'e copy paste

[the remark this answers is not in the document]
wagoodman: Also, this is the commit where I implemented the generic approach https://github.com/anchore/stereoscope/pull/154/commits/033d3e4cc97730d10c5001cb58d538e9ac88cc3e ... one of the tests fails in go 1.19 but passes in 1.20

pkg/file/path_set.go
```diff
@@ -1,20 +1,77 @@
...
 
+import (
+	"sort"
+)
+
 type PathSet map[Path]struct{}
 
-func NewPathSet() PathSet {
-	return make(PathSet)
+func NewPathSet(is ...Path) PathSet {
```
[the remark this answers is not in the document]
wagoodman: yup, also covered in https://github.com/anchore/stereoscope/pull/154#discussion_r1099356580

pkg/filetree/search.go
```diff
@@ -0,0 +1,477 @@
...
+	cache[req] = cacheResult{
+		Paths: paths,
+		Error: err,
+	}
+
+	return paths, err
+}
+
+// nolint: funlen
+func (sc searchContext) _pathsToNode(fn *filenode.FileNode, observedPaths file.PathSet, cache map[cacheRequest]cacheResult) (file.PathSet, error) {
```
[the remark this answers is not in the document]
wagoodman: probably more of a alex-ism

.github/workflows/validations.yaml
```diff
@@ -16,79 +16,28 @@ on:
       - main
   pull_request:
```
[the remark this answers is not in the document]
wagoodman: thanks -- I needed to add a new step here and decided to at least rebase what was here with the latest in syft

internal/string_set.go
```diff
@@ -0,0 +1,76 @@
+package internal
```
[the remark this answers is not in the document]
wagoodman: the goal was to use the generic implementation here, but yeah, this original one was written before we found the scylladb/go-set repo that we use heavily in syft. After the generic implementation didn't work I ended up just reverting to the "improved" state but didn't go forward with any full replacements (like scylladb/go-set).

pkg/image/image_test.go
wagoodman: Since this is with struct instances and am not planning in testing the json representation of these I think something like cmp.Diff is the right tool here.

.github/workflows/benchmark-testing.yaml
```diff
@@ -0,0 +1,58 @@
+name: "Benchmark testing"
```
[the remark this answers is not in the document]
wagoodman: > Do we want to add [podman] to bootstrap dependencies?

That would be a very heavy dependency to add, I think I'd prefer an update to the DEVELOPING.md on specifics here (just added in 4f28a64f59b4b3a14c1c7cde85d7e11083bbfa9a). 

> The `set-output` command is deprecated and will be disabled soon. 

righto -- I think we can add a dependabot for go and github actions for this kinda stuff (in another PR).

DEVELOPING.md
```diff
@@ -0,0 +1,52 @@
+# Developing
+
+## Getting started
+
+In order to test and develop in this repo you will need the following dependencies installed:
+- Golang
```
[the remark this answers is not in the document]
wagoodman: indeed -- this is where I added the note that it's at least required.

DEVELOPING.md
```diff
@@ -0,0 +1,52 @@
...
+- `filetree.FileTree`: a tree representing a filesystem. All nodes represent real paths (paths with no link resolution anywhere in the path) and are absolute paths (start with / and contain no relative path elements [e.g. ../ or ./]). This represents the filesystem structure and each node has a reference to the file metadata for that path.
+- `file.Reference`: a unique file in the filesystem, identified by an absolute, real path as well as an integer ID (`file.ID`s). These are used to reference concrete nodes in the `filetree.FileTree` and `image.FileCatalog` objects.
+- `file.Index`: stores all known `file.Reference` and `file.Metadata`. Entries are indexed with a variety of ways to provide fast access to references and metadata without needing to crawl the tree. This is especially useful for speeding up globbing.
+- `image.FileCatalog`:  an image-aware extension of `file.Index` that additionally relates `image.Layers` to `file.IDs` and provides a content API for any files contained within the image (regardless of which layer or squashed representation it exists in). 
+
+### Searching for files
```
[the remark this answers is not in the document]
wagoodman: thanks!!

pkg/file/reference_test.go
```diff
@@ -0,0 +1,279 @@
...
+				Reference:   &Reference{RealPath: "/another/place"},
+				LinkResolutions: []Resolution{
+					{
+						RequestPath: "/home",
+						Reference:   &Reference{RealPath: "/home"},
+					},
+				},
+			},
+			want: []Path{
+				"/home",
```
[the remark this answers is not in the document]
wagoodman: I don't think I understand -- `Path` is of type `string`, so there isn't a struct involved on the line in question, and most (all?) of the struct instances in testing use named fields.

Do you have another line you meant to highlight?

pkg/file/resolution.go
wagoodman: exactly 👍

pkg/image/image.go
wagoodman: > Should i.SquashedTreee() call i.squash()?

no -- squash is an expensive operation, we do this on Read() only and used the cached squashed trees for later usage.

> Does squash() affect fileCatalog?

no it should not -- the squash only makes structural changes / additions, but does not create any new files, so the catalog doesn't get modified.

pkg/image/layer.go
wagoodman: indeed, this is an important change. `Layer.tree` has been changed from `*filetree.FileTree` into a `filetree.Reader`, so no mutations are allowed. This change allows us to have a local instance that is mutable for indexing during the Read() call, but only persist and object that is read-only.

test/integration/fixture_image_simple_test.go
wagoodman: good question -- I think I'll keep it for now

test/integration/fixture_image_simple_test.go
wagoodman: I think I'll keep it for now

test/integration/fixture_image_simple_test.go
wagoodman: I think I'll keep it for now

.github/scripts/build.sh
```diff
@@ -0,0 +1,82 @@
...
+            CMD="GOARM=${GOARM} GOOS=${GOOS} GOARCH=${GOARCH} go build $FLAGS -o ${BIN_FILENAME} ${BUILD_TARGET}"
+            echo "${CMD}"
+            eval "${CMD}" || FAILURES="${FAILURES} ${GOOS}/${GOARCH}${GOARM}"
+        done
+    else
+        # Build non-arm here
+        if [[ "${GOOS}" == "windows" ]]; then BIN_FILENAME="${BIN_FILENAME}.exe"; fi
+        CMD="GOOS=${GOOS} GOARCH=${GOARCH} go build $FLAGS -o ${BIN_FILENAME} ${BUILD_TARGET}"
+        echo "${CMD}"
+        eval "${CMD}" || FAILURES="${FAILURES} ${GOOS}/${GOARCH}"
```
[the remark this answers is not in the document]
wagoodman: in theory the CI build cache should be reusing go build cache now, so what is usually 15 minutes should be taking something like 30 seconds. I can look into how to make this faster.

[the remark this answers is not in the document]
wagoodman: I've restricted the matrix build to have fewer cases which cuts it in half (down to 7 minutes). I think this is alright for now, but I can try to make this faster after this PR... is that a good compromise?

spiffcs (approved):

pkg/file/metadata.go
```diff
@@ -68,3 +88,40 @@ func NewMetadataFromSquashFSFile(path string, f *squashfs.File) (Metadata, error
...
+				if err := f.Close(); err != nil {
+					log.Warnf("unable to close file while obtaining metadata: %s", path)
+				}
+			}()
+		}
+
+		mimeType = MIMEType(f)
+	}
+
+	// TODO: should we clean up path to be the real, absolute path?
```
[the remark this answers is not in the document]
wagoodman: I think it is, let me get this change in

PR #158, opened 2023-02-21
Fix link cycle detection

wagoodman:
PR https://github.com/anchore/stereoscope/pull/154 added a regression where link resolution may result in falsely claiming that there is a link cycle for links that reference the same path twice while resolving links for a single path.

For example:
```
# links:
   /usr/local/bin/ksh -> /bin/ksh
   /bin -> /usr/bin
   /etc/alternatives/ksh -> /bin/ksh93

# real file: 
   /usr/bin/ksh93
```

In this case while resolving links in the `/usr/local/bin/ksh` path, the `/usr/bin` path is resolved twice, which under the current (buggy) implementation will falsely claim that there is a cycle relating to `/usr/bin`.

Relates to https://github.com/anchore/syft/issues/1586

### DIFF
--- a/pkg/file/path_set.go
+++ b/pkg/file/path_set.go
@@ -90,7 +90,7 @@ func (s PathCountSet) Add(ids ...Path) {
 			s[i] = 1
 			continue
 		}
-		s[i] += 1
+		s[i]++
 	}
 }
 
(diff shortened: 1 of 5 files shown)


kzantow (approved):
👍

spiffcs (approved):

Issue #132, opened 2022-06-23
Cleanup temp directories across image providers


wagoodman:
**What happened**:
_See docker/sbom-cli-plugin#22 for the original problem_; invocations of `docker sbom` (which uses the docker daemon provider) are not cleaning up temp files.

**What you expected to happen**:
All temp directories should be cleaned up.

**How to reproduce it (as minimally and precisely as possible)**:
`docker sbom alpine:latest` and `ls` the temp directory to see that there are `sbom-cli-plugin-*` directories still left (`docker sbom` uses stereoscope).

**Anything else we need to know?**:
 Today the docker daemon provider uses the tarball provider, each create sibling temp directories under a common root directory. We should be cleaning up the `image.tar` as soon as we create and provide an image from the tarball provider (from within the docker daemon provider --downside is that multiple calls to provide() will fail then [which hints that another solution may be needed here] ).

ltsonov-cb:
+1 on this - we noticed abnormal disk usage when utilizing syft and narrowed down the issue to this root cause as well. 
The code to work around it is a bit clunky as users of our wrapper APIs expect that `image.Cleanup()` is enough - but that leaves .tar files hanging around.

slimdevl:
I think this is an interesting problem. There are a few options here:

* option 1: have the file.TempFileGenerator remember generators and call their cleanup method when the top level cleanup is called.... https://github.com/anchore/stereoscope/blob/main/pkg/file/temp_dir_generator.go#L16
* option 2: provide a generic cleanup array on the image which can have methods added by any provider.
* option 3: provide a Cleanup interface method for a provider - and have the provider call the cleanup. (this feels off since the client.go file is the one allocating the temp generator)

Recommendations from the team - I am happy to open a PR with a little guidance ☺️

slimdevl:
> I think this is an interesting problem. There are a few options here:
> 
> * option 1: have the [file.TempFileGenerator](https://github.com/anchore/stereoscope/blob/main/client.go#L21)  remember generators and call their cleanup method when the top level cleanup is called.... https://github.com/anchore/stereoscope/blob/main/pkg/file/temp_dir_generator.go#L16
> * option 2: provide a generic cleanup array on the image which can have methods added by any provider.
> * option 3: provide a Cleanup interface method for a provider - and have the provider call the cleanup. (this feels off since the [client.go](https://github.com/anchore/stereoscope/blob/main/client.go#L103) file is the one allocating the temp generator)
> 
[...]

PR #161, opened 2023-02-27
fix directory leak leading to out of disk

slimdevl:
### [Issue 132](https://github.com/anchore/stereoscope/issues/132)

# Why
When using stereoscope as a library, for long running processes, the relationship between the providers may cause files in temp to not get cleaned up after `image.Image.Cleanup` is called. 

# What
* enable `image.Image.Cleanup` to own `tmpDirGenerator` and perform cleanup any directories related to the image.

# How Tested
* per guide
```
make bootstrap
make unit
make integration
```

* using the basic example
```
$ docker pull ruby:2.6-alpine3.11
$ ls -l /tmp/ | grep stereo
$ go run basic.go docker:ruby:2.6-alpine3.11 &>/dev/null
$ ls -l /tmp/ | grep stereo
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -170,5 +170,5 @@ func SetBus(b *partybus.Bus) {
 // Cleanup deletes all directories created by stereoscope calls. Note: please use image.Image.Cleanup() over this
 // function when possible.
 func Cleanup() {
-	log.Warn("cleanup is deprecated - no need to call. please use image.Image.Cleanup()")
+	log.Debug("cleanup is deprecated - no need to call. please use image.Image.Cleanup()")
 }
--- a/client.go
+++ b/client.go
@@ -18,8 +18,6 @@ import (
 	"github.com/wagoodman/go-partybus"
 )
 
-var rootTempDirGenerator = file.NewTempDirGenerator("stereoscope")
-
 func WithRegistryOptions(options image.RegistryOptions) Option {
 	return func(c *config) error {
 		c.Registry = options
@@ -100,7 +98,7 @@ func GetImageFromSource(ctx context.Context, imgStr string, source image.Source,
 
 func selectImageProvider(imgStr string, source image.Source, cfg config) (image.Provider, error) {
 	var provider image.Provider
-	tempDirGenerator := rootTempDirGenerator.NewGenerator()
+	tempDirGenerator := file.NewTempDirGenerator("stereoscope")
 	platformSelectionUnsupported := fmt.Errorf("specified platform=%q however image source=%q does not support selecting platform", cfg.Platform.String(), source.String())
 
 	switch source {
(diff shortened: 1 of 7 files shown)


client.go
```diff
@@ -172,7 +170,5 @@ func SetBus(b *partybus.Bus) {
 // Cleanup deletes all directories created by stereoscope calls. Note: please use image.Image.Cleanup() over this
 // function when possible.
 func Cleanup() {
-	if err := rootTempDirGenerator.Cleanup(); err != nil {
-		log.Errorf("failed to cleanup tempdir root: %w", err)
-	}
+	log.Warn("cleanup is deprecated - no need to call. please use image.Image.Cleanup()")
```
kzantow, line 173: This wouldn't need to be a `WARN` level message

kzantow:
Are you able to sign-off your commits? https://github.com/anchore/stereoscope/blob/main/CONTRIBUTING.md#sign-off-your-work

client.go
slimdevl: Ok, I wasn't sure. I can modify that 👍🏻

kzantow:
@slimdevl I'm not entirely sure why DCO isn't passing, it _looks_ right. Did you use `git commit -s` / `git commit --signoff`? Maybe try the steps listed here?: https://github.com/anchore/stereoscope/pull/161/checks?check_run_id=11626027320

slimdevl:
> 'm not entirely sure why DCO isn't passing, it _looks_ right. Did you use

I manually added those fields as recommended by the guide. Let me see if I can get the `-s` (maybe there is something I need to figure out there)

slimdevl:
> @slimdevl I'm not entirely sure why DCO isn't passing, it _looks_ right. Did you use `git commit -s` / `git commit --signoff`? Maybe try the steps listed here?: https://github.com/anchore/stereoscope/pull/161/checks?check_run_id=11626027320

This worked like magic, thanks @kzantow !

client.go
```diff
@@ -172,7 +170,5 @@ func SetBus(b *partybus.Bus) {
 // Cleanup deletes all directories created by stereoscope calls. Note: please use image.Image.Cleanup() over this
 // function when possible.
 func Cleanup() {
-	if err := rootTempDirGenerator.Cleanup(); err != nil {
-		log.Errorf("failed to cleanup tempdir root: %w", err)
-	}
+	log.Debug("cleanup is deprecated - no need to call. please use image.Image.Cleanup()")
```
wagoodman, line 173: Deprecation shouldn't change behavior (this would be a breaking change). Instead we should up date the comment.

wagoodman (approved):
Great find! I think promoting the temp dir generator to the image makes sense. I've pushed a few of adjustments:
- update the comment for the package Cleanup() function to clearly indicate that this is deprecated
- restore the root temp directory generator (to facilitate the package Cleanup function)
- Added both removal of the contentCacheDir and cleanup of the tmp directory generator in the Image object. This way someone constructing their own image object could in theory provider a temp dir generator that did not create the contentCacheDir and everything would still work.

PR #164, opened 2023-03-22
Add ability to explicitly specify an Authenticator as well as Keychain.

vaikas:
It would be nice to be able to specify the interfaces defined in the GGCR (for authn.Authenticator and authn.Keychain). Currently, the types are structs and it's not possible to wire in things like Keychains, or explicit Authenticators

Authenticator: for cases where you know the particular image you are referring to as well as which authenticator to use)
Keychain: for cases where do you not necessarily know, but want to be able to support multiple clouds helpers

This PR is just making the surgical changes to add them into the underlying structs to be usable. I could see in the future, then using them, so that they could take advantage of multikey chains like `ko` and `cosign` here:

https://github.com/ko-build/ko/blob/4e6b9c9a67718568bcd2e53f33a9787573b29741/pkg/commands/config.go#L48-L54
https://github.com/sigstore/cosign/blob/a00174fc6abef6143d35271ea4a3bddf1c7ceba5/cmd/cosign/cli/options/registry.go#L98-L106

For the life of me I can not figure out a way to test the function to validate the RemoteOptions are what they should be because they are functions, and the arguments to functions are private, so I can't even call the function and observe the side effects. But added tests where I could :)

### DIFF
--- a/pkg/image/oci/registry_provider.go
+++ b/pkg/image/oci/registry_provider.go
@@ -111,11 +111,15 @@ func prepareRemoteOptions(ctx context.Context, ref name.Reference, registryOptio
 	}
 
 	// note: the authn.Authenticator and authn.Keychain options are mutually exclusive, only one may be provided.
-	// If no explicit authenticator can be found, then fallback to the keychain.
+	// If no explicit authenticator can be found, check if explicit Keychain has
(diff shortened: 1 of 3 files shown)


vaikas:
As an extra special bonus, thank you @wagoodman for picking this up in syft also, much appreciated!!!

PR #165, opened 2023-03-29
feat: add ordered preference list of image source location for DetermineDefaultImagePullSource

spiffcs:
## Enable upstream tooling more pull source configuration

Enables: https://github.com/anchore/grype/issues/1204

Currently syft (and by dependency, grype) follow a local out pattern of checking for a source for a given input:

Current state:
EX: `alpine:latest` would check the Docker Daemon, then the Podman Daemon, then default to returning `OciRegistrySource` to the caller if connectivity to the daemons could not be established.

This update allows the caller more preference in checking if they would prefer to default to just shortcutting to finding an image for a registry.

## For discussion
NOTE: this implementation previously did NOT check if the image was on the registry before returning `OCIRegistry` as the preferred source. If the caller puts registry as the 0 indexed preference, then that is the returned source. 

If there is an error upstream the caller should decide if it wants to come back and attempt local daemon sources before erroring out.

We have to go all the way to this location before the returned `Source` is used in syft to 
https://github.com/anchore/syft/blob/2fa238af7ca9e64c3553b83481a3d05a3c9cdaaf/syft/source/source.go#L160

### DIFF
--- a/pkg/image/source.go
+++ b/pkg/image/source.go
@@ -11,8 +11,10 @@ import (
 	"time"
 
 	"github.com/anchore/stereoscope/internal/docker"
+	"github.com/anchore/stereoscope/internal/log"
 	"github.com/anchore/stereoscope/internal/podman"
 	"github.com/anchore/stereoscope/pkg/file"
+	"github.com/docker/docker/client"
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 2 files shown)


spiffcs:
This was implemented up in syft rather than at the stereoscope level:
https://github.com/anchore/syft/pull/1703

PR #166, opened 2023-03-31
Preserve time and expose link strategy

wobito:
This PR does two things:

1. Set the `Metadata` to hold the file `ModTime` and other time data when available.
2. expose the `LinkResolutionStrategy` to the walk callback via the `WalkConditions`

The reason for this PR is I'm using stereoscope to parse the squashedtree to a single tar file, and was required to make these little modifications in order to make that process work.

What do you think?


pkg/filetree/depth_first_path_walker.go
```diff
@@ -31,6 +31,8 @@ type WalkConditions struct {
 	// Whether we should consider children of this Node to be included in the traversal path.
 	// Return true to traverse children of this Node.
 	ShouldContinueBranch func(file.Path, filenode.FileNode) bool
+
+	LinkStrategy *LinkResolutionStrategy
```
wagoodman, line 35: I don't think this is a strong enough of a case to expose the LinkResolutionStrategy struct. That is, allowing overriding of `FollowAncestorLinks` is not advised and may result in potentially odd results. Instead I feel this should use the existing API surface area:
```suggestion
	LinkOptions []LinkResolutionOption
```
... and later craft a strategy internally
```golang
...
	linkOpts := []LinkResolutionOption{followAncestorLinks}
	linkOpts = append(linkOpts, w.conditions.LinkOptions...)
	linkStrat := newLinkResolutionStrategy(linkOpts...)
...
```

pkg/file/tarutil_test.go
```diff
@@ -93,6 +94,9 @@ func TestMetadataFromTar(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			f := getTarFixture(t, test.fixture)
 			metadata, err := MetadataFromTar(f, test.name)
+			metadata.ModTime = time.Time{}
```
wagoodman, line 97: nit: can you add non-empty values to ensure the fields are getting mapped?

pkg/filetree/depth_first_path_walker.go
wobito: Sweet!!! I'll make some adjustments. Thanks for the review :D

PR #170, opened 2023-04-12
Add format make target

wagoodman:
Splits the format-related tasks in the `lint-fix` target into a new `format` target (which `lint-fix` depends on).

Additionally updates the linter, lint targets, and thus there are code changes due to new linting behaviors. (The `Makefile` is the main file for review)


kzantow (approved):
`gosimports`-ed, nice!

PR #172, opened 2023-04-13
refactor: embed fs.FileInfo within file.Metadata

### LINKED ISSUE #171: Embed fs.FileInfo in Metadata Struct?

I'd like a way to obtain a (or ideally, preserve the original) [fs.FileInfo](https://pkg.go.dev/io/fs#FileInfo) associated with a given [file.Metadata](https://pkg.go.dev/github.com/anchore/stereoscope/pkg/file#Metadata) value. `ModTime`/`AccessTime`/`ChangeTime` were added directly to the `Metadata` struct in #166, which works, but that approach requires modifying the `Metadata` struct each time a new field is needed, and then wiring up the various sources that Stereoscope supports (TAR, Singularity/SquashFS, OS filesystem.)

I suspect it might be a better solution to simply embed an actual `fs.FileInfo` obtained from the source within the `Metadata` struct? This would allow anyone with a `Metadata` to get any metadata supported by the `io/fs` package, without requiring additional fields to be added to the `Metadata` struct each time. I believe it'd also allow removal of some of the source-specific handling.

**Why is this needed**:

I'm looking to get access to additional fields available from the source `fs.FileInfo` (ex. `Xattrs`).

**Additional context**:

I'd be willing to do the leg work on a PR for this, if the idea is sound. Thanks!

tri-adam:
As discussed in today's community call... here is a proposal to go with #171. Thanks in advance for takin the time to discuss it, and for any feedback on this change.

Embed the original `fs.FileInfo` directly within the `file.Metadata struct`, replacing the previous `Size`, `IsDir`, `Mode`, and `ModTime` fields.

I'm also proposing the removal of the recently added `AccessTime` and `ChangeTime` fields from the `file.Metadata` struct, with the argument that they're currently unset except when the `Metadata` refers to a TAR layer, and the layer uses a format other than the relatively common `USTAR` ([ref](https://pkg.go.dev/archive/tar#Format)). Technically, a `fs.FileInfo` still exposes these via `fs.FileInfo.Sys()` if the returned value is of type `*tar.Header`. I believe this approach discourages use of fields which won't be set very often, while still making them available to those willing to dig. All that being said, I'm not strongly opinionated here!

Closes  #171

### DIFF
--- a/pkg/file/metadata.go
+++ b/pkg/file/metadata.go
@@ -7,11 +7,14 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"time"
 
 	"github.com/sylabs/squashfs"
 
 	"github.com/anchore/stereoscope/internal/log"
 )
+
+var _ fs.FileInfo = (*ManualInfo)(nil)
 
 // Metadata represents all file metadata of interest.
 type Metadata struct {
@@ -25,6 +28,38 @@ type Metadata struct {
 	GroupID         int
 	Type            Type
 	MIMEType        string
+}
+
+type ManualInfo struct {
+	NameValue    string
+	SizeValue    int64
+	ModeValue    fs.FileMode
+	ModTimeValue time.Time
+	SysValue     any
+}
+
+func (m ManualInfo) Name() string {
+	return m.NameValue
+}
+
+func (m ManualInfo) Size() int64 {
+	return m.SizeValue
+}
+
+func (m ManualInfo) Mode() fs.FileMode {
+	return m.ModeValue
+}
+
+func (m ManualInfo) ModTime() time.Time {
+	return m.ModTimeValue
+}
+
+func (m ManualInfo) IsDir() bool {
+	return m.ModeValue.IsDir()
+}
+
+func (m ManualInfo) Sys() any {
+	return m.SysValue
 }
 
 func NewMetadata(header tar.Header, content io.Reader) Metadata {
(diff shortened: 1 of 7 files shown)


wagoodman:
I agree with your point about AccessTime and ChangeTime they should probably be removed (though still accessible as you mentioned... CC: @wobito ).

I've got a draft branch in syft that uses these changes. I've made a few adjustments to this branch based on using it:

- Adding a `file.ManualInfo` that acts like your `mockFileInfo` is fairly useful not only for testing in this repo, but also in syft for decoding from SBOM documents and creating `file.Metadata` instances, which won't have access to the original `FileInfo` objects, but have much of the static information available. I was tempted to put this in Syft since it seems like purely a syft concern until I noticed the similarity to the mock object (so was swayed to put it in stereoscope instead).
- It seems that being able to compare `file.Metadata` objects is needed both in stereoscope and syft, so adding a simple `Equal(other Metadata) bool` function helped in testing and for usage up in syft without the caller needing to worry about the implementation details.

How do you feel about these changes? Mind if I push on this branch @tri-adam ?

tri-adam:
Thanks for the review @wagoodman!

> I've got a draft branch in syft that uses these changes. I've made a few adjustments to this branch based on using it:
> 
> * Adding a `file.ManualInfo` that acts like your `mockFileInfo` is fairly useful not only for testing in this repo, but also in syft for decoding from SBOM documents and creating `file.Metadata` instances, which won't have access to the original `FileInfo` objects, but have much of the static information available. I was tempted to put this in Syft since it seems like purely a syft concern until I noticed the similarity to the mock object (so was swayed to put it in stereoscope instead).

Sounds like a great idea, yeah.

> * It seems that being able to compare `file.Metadata` objects is needed both in stereoscope and syft, so adding a simple `Equal(other Metadata) bool` function helped in testing and for usage up in syft without the caller needing to worry about the implementation details.

That makes sense too. Same pattern as [(time.Time).Equal(t)](https://pkg.go.dev/time#Time.Equal) and for similar reasons.

> How do you feel about these changes? Mind if I push on this branch @tri-adam ?

No objections here! Let me know if I can help out with anything...

wagoodman (approved):

PR #173, opened 2023-04-21
feat: use credential helpers with the default keychain

ivanstanev:
This allows tools like Syft to automatically pick up and provision credentials when pulling from a Container Registry. This is also useful for running Syft inside Kubernetes and automatically grab the necessary credentials from the cloud provider.


ivanstanev:
@wagoodman sounds like the better approach!

PR #174, opened 2023-05-01
feat: Add containerd support

### LINKED ISSUE #67: Add containerd support
In the same vein of supporting other container runtimes, it would be ideal to support pulling images from containerd directly.

shanedell:
feat: Add containerd support

Closes #67

**NOTES**:

- Couldn't quite figure how to get `containerd` to use the party bus and pull event properly.
- Currently the display for pulling an image is the same as when you run `ctr image pull`.
- Use environment variables for some of the auth configuration, possibly better way to do it?
- Currently removed the pull_status.go file since it wasn't being used.

@wagoodman Any tips on integrating the pull_status and party bus?

Currently output looks something like this:

```bash
>  go run cmd/syft/main.go containerd:docker.io/library/python:3.7

index.docker.io/library/python:3.7:                                               resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:741bcc4e5900df61be4d3c1061ec1623074b50cce1a72269ffa87b9457406392:    done           |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:e1386c79b05d181ea7a1a705be8caa5f90fd09302362184c96388c98b3ba8190: done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:5c75a57b205e0e5bc8201af9d669b1dcd96e655203943f6732262261a9675723:    done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:c70a645202fbae11b330afc0fbc4875790df7fa4cbd2e817a30a8017db1b9726:   done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:b0248cf3e63c73d0e496a67807d056ca41d5e968b61087e8eca2cf4b9b4d7b99:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:127e97b4daf784e08840a21765f0d4f251192ef2994d0e4a253490f81e63955b:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:2d62772179761f8fddfaed03ed2bdf7078b103b193d18d79a9b49364830d56cf:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:273fcda609d894158ae396f62b319d33f7b02266fbf6bd88c4950771f1a40a83:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:713e0eee7be8734b79f270d5f920eaab5331b1b206b16c05e42a3ad75b2e7321:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:760b56890cca3bf0ac9cc5e1b193cdc449ade3de3ec5bcc1825ff61f676ddf2f:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:0336c50c9f6942b660e433b1086238eec37057c34b14c4e3b28bd7bf05bd84ba:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:1b89f3c7f7da8adf032a33a75d1b659cee33179ecb88ea0ba75e4fc58ebe63a6:    done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 0.4 s                                                                    total:   0.0 B (0.0 B/s)                                         
 ✔ Parsed image            
 ✔ Cataloged packages      [432 packages]
NAME                          VERSION                         TYPE   
adduser                       3.118                           deb     
apt                           2.2.4                           deb     
autoconf                      2.69-14                         deb     
automake                      1:1.16.3-2                      deb     
autotools-dev                 20180224.1+nmu1                 deb     
base-files                    11.1+deb11u6                    deb     
base-passwd                   3.5.51                          deb
```


Shanedell:
@wagoodman I pushed an update to fix the static analysis. I think the test may still fail, any idea on why that is?

wagoodman:
@Shanedell this is awesome! I tried to answer your question via comments, but needed to learn more about the containerd API to respond meaningfully. In exploring the API I eventually discovered one way to get the pull status from the API and had most of it done... so I pushed that work to your branch (My apologies if this was something you were looking forward to implement yourself... feel free to change/delete anything I pushed to your branch).

I also pushed a branch in syft that uses the events published in stereoscope to show the containerd pull status https://github.com/anchore/syft/compare/containerd-ui-handlers (this was mostly a copy and paste of the existing docker pull status handler).

So, now that I've done this entirely backwards, let me try to answer your first question (how to use partybus in the context of the containerd API):

The bus used in syft/stereoscope/grype can be thought of like a logger, but instead for rich objects which listeners can use dynamically. Events should be read-only objects that give callers observability into some notable event that is happening "under the hood" of a lib/application. The containerd API seems pretty powerful, but at the same time, I feel that the caller has to do a lot of work to get status for a pull of a single image... this is where most of the complexity came in. The `PullStatus` represents the read-only object that is published on the bus while the `apiState` encapsulated within it is a facade wrapping the complications of the containerd API. 

I couldn't figure a way to "stream" events back for progress, which means that there needed to be something that polled the API for download progress. This could have been driven by callers on PullStatus, but that would break the rule of thumb that the objects on the bus are read only... that is, if calling `PullStatus.Current()` (or something similar to get the current download status) made containerd API calls, that would be surprising and most likely bad.  There is a RWMutex to coordinate the state being read by `PullStatus` and written by `apiState` to minimize data race situations.

Checkout the reference implementation I pushed and shout out if you have questions.

wagoodman:
The test failure you're seeing is due to a new source being added (containerd) but not being tested under an existing test that expects to see coverage for all currently implemented sources: https://github.com/anchore/stereoscope/actions/runs/4909252292/jobs/8765546160#step:11:702
```
    fixture_image_symlinks_test.go:81: probably missed a source during testing, double check that all image.sources are covered
```
The `TestImageSymlinks` test requires one more test case in the test table:
```
		{
			name:   "FromContainerd",
			source: "containerd",
		},
```

This may require CI changes as well as documented assumptions for development in the DEVELOPING.md file too. Shout out if you'd like more help with this failure.

internal/containerd/client.go
```diff
@@ -0,0 +1,18 @@
+package containerd
+
+import (
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/defaults"
+)
+
+var (
+	defaultAddr = defaults.DefaultAddress
```
wagoodman, line 9: does this address change with a rootless configuration? (say somewhere under `/run/user/$(id -u)/containerd...`)

Shanedell:
@wagoodman I like the change, I pulled them locally and tested and everything is working well. I appreciate the help and the information. I will fix the test and will figure out a way to allow for rootless containerd.

internal/containerd/client.go
Shanedell: No it does not. It always tries to use `/var/run/containerd/containerd.sock`. However, when I manually updated the file to point to the `api.sock` for rootless it still worked fine so I just to find a way to wire that in.

Shanedell: I take this back actually, seems it may not work with that. From what I know of containerd's command line tool `nerdctl` they also use that same `defaults.DefaultAddress`. Inside of the docs, [rootless containerd](https://github.com/containerd/containerd/blob/main/docs/rootless.md), it doesn't specify a different address in the `nerdctl` command so it would use that default value still. So I assume for root and rootless this works maybe?

Shanedell: I think I have something working for this. I plan to make a PR to your syft branch for adding the containerd address flag. The way I am currently passing the value from `syft` to `stereoscope` is through setting a value in the context, would it be preferred that I send it via a parameter instead? In `syft` I had to add `containerdAddress string` to bunch of methods. The only way I could find getting around this was adding it as an `Application` attribute. However, when saving to a file then you would be get this value mapped to your file, which if we want is fine but I figured we wouldn't that in an output file.

Shanedell: I pushed up a change to that should fix the integration tests. Also allows the passing of the `containerdAddress` to different functions so the client which use that address if specified otherwise will try to use the default address. There will be a matching PR for `syft`

Shanedell:
@wagoodman So I am fixing the integration test issue but this requires containerd building an image, which it can't do. So my idea was assume docker is available also. So then docker builds the image, we save it off, load into containerd, remove save image. However, with this there's another issue, this being that when the provider gets sent the image it does not include the registry host. This make containerd throw an error because it thinks the tag is the host port.

wagoodman:
I'm currently looking into an oddity:
```bash
# run the tests...
runner@fv-az459-985:~/work/stereoscope/stereoscope$ go test ./test/integration/platform_test.go
--- FAIL: TestPlatformSelection (7.98s)
    --- FAIL: TestPlatformSelection/ContainerdDaemonSource/linux/arm64 (0.89s)
        platform_test.go:109:
                Error Trace:    /home/runner/work/stereoscope/stereoscope/test/integration/platform_test.go:109
                Error:          Received unexpected error:
                                unable to use ContainerdDaemonSource source: unable to save image tar for image="index.docker.io/library/busybox:1.31": content digest sha256:fd4a8673d0344c3a7f427fe4440d4b8dfd4fa59cfabbd9098f9eb0cb4ba905d0: not found
                Test:           TestPlatformSelection/ContainerdDaemonSource/linux/arm64
    --- FAIL: TestPlatformSelection/ContainerdDaemonSource/linux/amd64 (0.00s)
        platform_test.go:109:
                Error Trace:    /home/runner/work/stereoscope/stereoscope/test/integration/platform_test.go:109
                Error:          Received unexpected error:
                                unable to use ContainerdDaemonSource source: unable to save image tar for image="index.docker.io/library/busybox:1.31": content digest sha256:fd4a8673d0344c3a7f427fe4440d4b8dfd4fa59cfabbd9098f9eb0cb4ba905d0: not found
                Test:           TestPlatformSelection/ContainerdDaemonSource/linux/amd64
FAIL
FAIL    command-line-arguments  9.698s
FAIL

# take a look at the state containerd is left in... 
runner@fv-az459-985:~/work/stereoscope/stereoscope$ ctr i ls

ERRO[0000] failed calculating size for image index.docker.io/library/busybox:1.31  error="content digest sha256:fd4a8673d0344c3a7f427fe4440d4b8dfd4fa59cfabbd9098f9eb0cb4ba905d0: not found"

REF                                  TYPE                                                      DIGEST                                                                  SIZE    PLATFORMS                                                                                                            LABELS
index.docker.io/library/busybox:1.31 application/vnd.docker.distribution.manifest.list.v2+json sha256:95cf004f559831017cdf4628aaf1bb30133677be8702a8c5f2994629f637a209 2.5 KiB linux/386,linux/amd64,linux/arm/v5,linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/mips64le,linux/ppc64le,linux/s390x -
```
I see this in CI https://github.com/anchore/stereoscope/actions/runs/6098722348/job/16549192993 and am going to attempt to get this replicated locally. It seems that using the containerd API we can get a pulled state where the store is not valid -- fun!

wagoodman:
There seems to be some sane defaults in containerd that cause issues when handling corner cases, such as multi-arch images and non linux/amd64 os/arch combos. I've added additional platform parsing logic to this branch. It came down to not really trusting the `containerd.Image` object returned from `client.GetImage()` -- it would provide a valid architecture in some cases, and no architecture in others, even when providing a digest to a specific image manifest (not a multiarch manifest). 

There are a few more things I'll try to add to this branch:
- [ ] more unit tests to cover platform handling
- [ ] try to parse additional tags from the underlying image (the current label approach was not yielding expected results)

test/integration/fixture_image_simple_test.go
```diff
@@ -119,6 +128,15 @@ func TestSimpleImage(t *testing.T) {
 
 	for _, c := range simpleImageTestCases {
 		t.Run(c.source, func(t *testing.T) {
+			if runtime.GOOS != "linux" {
```
wagoodman, line 131: I ran into issues testing linux-only sources so made adjustments to skip tests when it's not a valid environment

.github/workflows/validations.yaml
```diff
@@ -49,12 +49,15 @@ jobs:
 
   Integration-Test:
     name: "Integration tests"
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
wagoodman, line 52: note: ubuntu 22.04 was required to install `rootlesskit` for the rootless containerd setup, so I bumped this everywhere to be consistent

internal/podman/client.go
```diff
@@ -77,11 +77,20 @@ func ClientOverUnixSocket() (*client.Client, error) {
...
-		log.Debugf("no socket address was found. Trying default address: %s", socketPath)
+
+		var socketPath string
+		uid := os.Getuid()
+		switch uid {
+		case 0:
+			socketPath = "/run/podman/podman.sock"
+		default:
+			socketPath = fmt.Sprintf("/run/user/%d/podman/podman.sock", os.Getuid())
+		}
```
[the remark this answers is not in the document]
wagoodman: broken out into a separate PR https://github.com/anchore/stereoscope/pull/196

pkg/image/image.go
```diff
@@ -49,8 +50,16 @@ func WithTags(tags ...string) AdditionalMetadata {
 		for _, t := range image.Metadata.Tags {
 			existingTags.Add(t.String())
 		}
+
 		for _, t := range tags {
-			tagObj, err := name.NewTag(t)
+			// it is possible that we are given references that have both a tag and a digest (or only one)
+			// we should only be allowing tags (stripping off digests if they are present)
+			fields := strings.Split(t, "@")
```
kzantow, line 57: would this be better to use `github.com/docker/distribution/reference` (or the other one) to parse the reference?

pkg/image/source.go
```diff
@@ -35,6 +37,7 @@ const SchemeSeparator = ":"
 
 var sourceStr = [...]string{
 	"UnknownSource",
+	"ContainerdDaemonSource",
```
kzantow, line 40: Should this just be `ContainerdDaemon` like `DockerDaemon` and `PodmanDaemon`?

kzantow (approved):
I have a general question about the environment variables being read: are these standard containerd variables? Should these instead be some sort of configuration that can be passed in?

LGTM otherwise, nothing blocking, just a few small comments. 👍

pkg/image/containerd/daemon_provider.go
```diff
@@ -0,0 +1,485 @@
...
+	}
+
+	useRegAuth, err := strconv.ParseBool(os.Getenv("USE_REGISTRY_AUTH"))
+	if err != nil {
+		useRegAuth = false
+	}
+
+	if useRegAuth {
+		username := os.Getenv("CTR_REGISTRY_USERNAME")
+		secret := os.Getenv("CTR_REGISTRY_PASSWORD")
```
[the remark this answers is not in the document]
wagoodman: the CONTAINERD_ADDRESS is, I'll double check on the rest

pkg/image/source.go
wagoodman: good find

pkg/image/containerd/daemon_provider.go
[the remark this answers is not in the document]
wagoodman: no these don't look standard -- @shanedell are these env var names used in other applications? Or were they made in this PR as a necessity?

internal/containerd/client.go
```diff
@@ -0,0 +1,58 @@
...
+	return client, nil
+}
+
+func Address() string {
+	envAddr := os.Getenv("CONTAINERD_ADDRESS")
+	if envAddr != "" {
+		return envAddr
+	}
+
+	if os.Getpid() == 0 {
```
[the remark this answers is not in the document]
wagoodman: You're not missing something -- this should be `uid` not `pid`. I'll take a second pass at this section and try and make the logic consistent with the recent podman refactor

pkg/image/containerd/daemon_provider.go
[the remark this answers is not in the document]
wagoodman: I haven't been able to find anything yet. It seems that the registry options map quite nicely onto the containerd options at first glance. I'll look into mapping these options and requiring that input for containerd API access.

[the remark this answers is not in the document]
shanedell: @wagoodman I can't currently remember where I got these environment variables from. I believe it should have from either the [containerd](https://github.com/containerd/containerd) repo or the [nerdctl](https://github.com/containerd/nerdctl) repo as I looked at these both extensively to figure out how things were working. I may have just made some of those up on the fly at the time to make sure the authentication could work and can be updated.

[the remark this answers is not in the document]
wagoodman: I wasn't able to reproduce using these env vars with nerdctl or ctr commands (image pull, login, etc). A quick search in each repo also didn't really show any equivalent env vars. I'm going to go the path of mapping this onto the existing `RegistryOptions`, since containerd is a more low-level tool it seems to require a lot of these options force fed as CLI options directly... which hints that it's really acting more like the OCI registry provider anyway.

I've got a mock up locally, I'll get it cleaned up and pushed tonight (maybe)

pkg/image/image.go
wagoodman: I think I'll leave this one as is to get it over the finish line since the only thing it's doing is removing a digest if provided (and ignoring if there is no tag).

wagoodman (approved):
awesome work @shanedell 🏅  I'll get your PR updated in syft

Issue #182, opened 2023-05-25
Goroutine Leak in Long-Running Services with Stereoscope


iAklis:
**What happened**:
I've noticed a goroutine leak when using stereoscope in a service that needs to run continuously. The issue seems to be with connections to the Docker daemon via the Unix socket not being released.

**What you expected to happen**:
After continuously communicating with the Docker daemon, Stereoscope should not leave long connections unreleased.

**How to reproduce it (as minimally and precisely as possible)**:

Here's a minimal reproducible example:

```go
package main

import (
    "context"
    "fmt"
    "net/http"
    _ "net/http/pprof"
    "os"
    "os/signal"
    "syscall"

    "github.com/anchore/stereoscope/internal/log"
    "github.com/anchore/stereoscope/pkg/image"

    "github.com/anchore/stereoscope"
)

func main() {
    go func() {
        http.ListenAndServe(":6060", nil)
    }()

    signalChan := make(chan os.Signal, 1)
    signal.Notify(signalChan, os.Interrupt, syscall.SIGTERM)

    go func() {
        for {
            ctx := context.TODO()

            ImageSource := image.DockerDaemonSource
            UserInput := "busybox:latest"
            img, err := stereoscope.GetImageFromSource(ctx, UserInput, ImageSource)
            if err != nil {
                panic(err)
            }
            cleanup := func() {
                if err := img.Cleanup(); err != nil {
                    log.Warnf("unable to cleanup image=%q: %w", UserInput, err)
                }
            }

            for _, layer := range img.Layers {
                fmt.Printf("layer: %s\n", layer.Metadata.Digest)
            }

            cleanup()

        }
    }()

    <-signalChan

}
```
**Anything else we need to know?**:

**Environment**:
```
OS: Linux kaze 5.15.0-71-generic #78-Ubuntu SMP Tue Apr 18 09:00:29 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
Docker: 20.10.21 # installed by `sudo apt install docker.io`
Stereoscope: v0.0.0-...-e14bc44 (Latest as of May 22, 2023)
```

kzantow:
Thanks for the report @iAklis. After looking at this issue a bit and your PR, it seems as though you've ascertained that the network connections are being held open, resulting in an increase in resource usage.

The current model creates a new docker client per `GetImageFromSource` calls.

[...]

PR #183, opened 2023-05-25
fix: goroutine leak with docker daemon 

iAklis:
This commit addresses a goroutine leak that occurs when continuously communicating with the Docker daemon via the Unix socket. Previously, these connections were not being properly released, leading to a growing number of goroutines over time.

The first solution with https://github.com/anchore/stereoscope/pull/183/commits/4cf42ace9aa21cc44fc18bab7721618fdf13bb73 involved modifying the http.Transport of the Docker client to disable keep-alives and set an idle connection timeout. This effectively closes idle connections, thereby preventing the goroutine leak.

Based off of [comments on the issue](https://github.com/anchore/stereoscope/issues/182#issuecomment-1634919267) this was pivoted to close the client after the provider is used for the first time.

Resolves: #182

### DIFF
--- a/client.go
+++ b/client.go
@@ -81,7 +81,10 @@ func GetImageFromSource(ctx context.Context, imgStr string, source image.Source,
 		}
 	}
 
-	provider, err := selectImageProvider(imgStr, source, cfg)
+	provider, cleanup, err := selectImageProvider(imgStr, source, cfg)
+	if cleanup != nil {
+		defer cleanup()
+	}
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


wagoodman:
Hey @iAklis I updated the PR based on feedback in the issue -- that being said, you can treat this as a suggested path... if you don't like this approach we can chat further. Does the branch with the current adjustments also fix the behavior you're seeing?

iAklis:
@wagoodman Yes, it works.

wagoodman (approved):

Issue #188, opened 2023-06-01
Platform selection in docker client has unnecessary error


willmurphyscode:
**What happened**:

```
❯ go run examples/basic.go docker:busybox:1.31@sha256:91c15b1ba6f408a648be60f8c047ef79058f26fa640025f374281f31c8704387
[0000] DEBUG image: source=DockerDaemon location=busybox:1.31@sha256:91c15b1ba6f408a648be60f8c047ef79058f26fa640025f374281f31c8704387
[0000] DEBUG pulling docker image="busybox:1.31@sha256:91c15b1ba6f408a648be60f8c047ef79058f26fa640025f374281f31c8704387"
[0000] DEBUG using docker config="/Users/willmurphy/.docker/config.json"
[0000] DEBUG using docker credentials for "index.docker.io/v1/"
panic: unable to use DockerDaemon source: image has unexpected architecture "s390x", which differs from the user specified architecture "arm64"

goroutine 1 [running]:
main.main()
        /Users/willmurphy/work/stereoscope/examples/basic.go:38 +0x60c
exit status 2
```

This error affects syft: 

```
❯ syft packages docker:busybox:1.31@sha256:91c15b1ba6f408a648be60f8c047ef79058f26fa640025f374281f31c8704387           
 ⠹ Pulling image           

2023/06/01 13:52:00 error during command execution: 1 error occurred:
        * failed to construct source from user input "docker:busybox:1.31@sha256:91c15b1ba6f408a648be60f8c047ef79058f26fa640025f374281f31c8704387": could not fetch image "busybox:1.31@sha256:91c15b1ba6f408a648be60f8c047ef79058f26fa640025f374281f31c8704387": scheme "docker" specified; image retrieval using scheme parsing (busybox:1.31@sha256:91c15b1ba6f408a648be60f8c047ef79058f26fa640025f374281f31c8704387) was unsuccessful: unable to use DockerDaemon source: image has unexpected architecture "s390x", which differs from the user specified architecture "arm64"; image retrieval without scheme parsing (docker:busybox:1.31@sha256:91c15b1ba6f408a648be60f8c047ef79058f26fa640025f374281f31c8704387) was unsuccessful: unable to determine image source to select platform
```

**What you expected to happen**:

Since I specified an exact digest, and didn't pass `--platform`, I didn't expect a validation error about the user requested platform.

**How to reproduce it (as minimally and precisely as possible)**:

[...]

PR #189, opened 2023-06-09
Change platform selection logic

willmurphyscode:
This PR a couple things going on:

- Fix https://github.com/anchore/stereoscope/issues/188, essentially by reverting https://github.com/anchore/stereoscope/pull/152.
- Partially re-implement https://github.com/anchore/stereoscope/pull/152 in order to avoid re-introducing https://github.com/anchore/stereoscope/issues/149
- Add integration tests to protect against regression of https://github.com/anchore/stereoscope/issues/188 or https://github.com/anchore/stereoscope/issues/149
- Add some tests that use `busybox:1.31` on the `linux/s390x` platform. Previously, tests only used `linux/amd64` (which is an architecture match for many modern laptops) and `linux/arm64` (which is an architecture match for M1 macs). Adding `linux/s390x` images lets us pin down the behavior of image pulling that is almost certainly cross-platform. In other words, if we only have `arm64` and `amd64` test cases, then those test cases flip-flop between cross-platform and same-platform depending on whether the contributor is on an M1 or not, but adding some tests that use a platform found in fewer development laptops gives some tests that are _always_ cross-platform.

I wanted to point out a couple things for discussion:

1. It's not 100% clear to me that https://github.com/anchore/stereoscope/issues/149 represented incorrect behavior. I could see an argument that making the `registry` provider imitate docker's platform selection was the wrong call.
2. This pull request could be 2 different ones, one to fix https://github.com/anchore/stereoscope/issues/188, and one to re-fix https://github.com/anchore/stereoscope/issues/149.
3. The function `defaultPlatformIfNil` intentionally swallows an error if it occurs. That's because we're trying to guess a default behavior here, and I don't think an attempt to infer the desired default should ever return an error.


test/integration/platform_test.go
```diff
@@ -28,6 +30,12 @@ func TestPlatformSelection(t *testing.T) {
 			os:             "linux",
 			expectedDigest: "sha256:1ee006886991ad4689838d3a288e0dd3fd29b70e276622f16b67a8922831a853", // direct from repo manifest
 		},
+		{
```
spiffcs, line 33: not: should we add a `name` field to these tests to give some context beyond that we're testing different combinations of arch/os

I think both of us took at look at this earlier and were confused or unsure about the comments `// from generated manifest` vs `direct from repo manifest`and what they added to the case

spiffcs (approved):

test/integration/platform_test.go
willmurphyscode: We added a top-level comment explaining how this test works instead.

PR #196, opened 2023-09-07
Fix Podman client address resolution

wagoodman:
This refactors some of the podman client address resolution logic such that:

- podman env vars override config files
- rootless socket is attempted first before trying the default socket
- uncomments and fixes the podman integration tests
- refactors the podman os actions to use the afero FS abstraction (makes testing easier for the new client tests)

### DIFF
--- a/test/integration/podman_test.go
+++ b/test/integration/podman_test.go
@@ -1,9 +1,20 @@
 package integration
 
-// This was commented out until we can confirm the new behavior of the github runner
-// tests started throwing "read: connection reset by peer" when connecting to ssh://root@localhost:2222/run/podman/podman.sock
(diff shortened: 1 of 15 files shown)


test/integration/podman_test.go
```diff
@@ -1,9 +1,20 @@
 package integration
 
-// This was commented out until we can confirm the new behavior of the github runner
-// tests started throwing "read: connection reset by peer" when connecting to ssh://root@localhost:2222/run/podman/podman.sock
-// we might need to think of another creative way to test this, but in the meantime it has been failing stereoscope builds
-/*
```
wagoodman, line 6: note that all of the below code was commented out on main (hard to tell in the diff)

spiffcs (approved):

PR #212, opened 2024-01-08
Drop github.com/pkg/errors

jonjohnsonjr:
Most of these uses seem unintentional, but I've rewritten the intentional uses using fmt.Errorf.


wagoodman (approved):
thanks for the cleanup 🙌

PR #217, opened 2024-02-01
feat: account for osx special case socket

spiffcs:
## Summary
Some users encounter an error when their docker desktop `DOCKER_HOST` is not at the default location of `/var/run/docker.sock`

In newer versions of docker desktop the default is for this option to be disabled and for a new location to be added for a local users docker socket. This new common path can be found at `user.HomeDir, "Library/Containers/com.docker.docker/Data/docker.raw.sock"` on `darwin` systems

## To Reproduce and assert the fix
On your local make sure the advanced settings of docker desktop is disabled for using `/var/run/docker.sock`.

The setting is called `Allow the default Docker socket to be used (requires password)` under the `Advanced` tab in settings.

Run this command where `SOME_IMAGE_ID` is an image local to your machine.
```
syft -q <SOME_IMAGE_ID>
```
You should see an output like this:
Screenshot 2024-02-01 at 12 41 41 AM

When syft imports stereoscope with this fix you will no longer see an error as it knows how to use the correct new default socket on OSX:
Screenshot 2024-02-01 at 12 44 26 AM

### DIFF
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -2,12 +2,12 @@ package docker
 
 import (
 	"context"
-	"os/user"
-	"path/filepath"
-	"runtime"
 	"fmt"
 	"net/http"
 	"os"
+	"os/user"
+	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/docker/cli/cli/connhelper"
@@ -69,7 +69,7 @@ func GetClient() (*client.Client, error) {
 		}
 
 		macOSSocketPath := filepath.Join(user.HomeDir, "Library/Containers/com.docker.docker/Data/docker.raw.sock")
-		clientOpts = append(clientOpts, client.WithHost("unix://" + macOSSocketPath))
+		clientOpts = append(clientOpts, client.WithHost("unix://"+macOSSocketPath))
 		dockerClient, err = client.NewClientWithOpts(clientOpts...)
 		if err == nil {
 			_, err := dockerClient.Ping(context.Background())
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -5,13 +5,13 @@ import (
 	"fmt"
 	"net/http"
 	"os"
-	"os/user"
 	"path/filepath"
 	"runtime"
 	"strings"
 
 	"github.com/docker/cli/cli/connhelper"
 	"github.com/docker/docker/client"
+	"github.com/mitchellh/go-homedir"
 )
 
 func GetClient() (*client.Client, error) {
@@ -78,12 +78,11 @@ func GetClient() (*client.Client, error) {
 func newClient(os string, opts ...client.Opt) (*client.Client, error) {
 	switch os {
 	case "darwin":
-		localUser, err := user.Current()
+		hDir, err := homedir.Dir()
 		if err != nil {
-			return nil, fmt.Errorf("failed to get current user: %w", err)
+			return nil, err
 		}
-
-		macOSSocketPath := filepath.Join(localUser.HomeDir, "Library/Containers/com.docker.docker/Data/docker.raw.sock")
+		macOSSocketPath := filepath.Join(hDir, "Library/Containers/com.docker.docker/Data/docker.raw.sock")
 		opts = append(opts, client.WithHost("unix://"+macOSSocketPath))
 	default:
 	}
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -21,7 +21,6 @@ func GetClient() (*client.Client, error) {
 	}
 
 	host := os.Getenv("DOCKER_HOST")
-
 	if strings.HasPrefix(host, "ssh") {
 		var (
 			helper *connhelper.ConnectionHelper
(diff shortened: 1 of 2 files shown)


internal/docker/client.go
```diff
@@ -48,10 +52,33 @@ func GetClient() (*client.Client, error) {
...
+	}
+
+	// If on macOS and the default Unix socket didn't work, try the macOS-specific path
+	if runtime.GOOS == "darwin" {
+		user, err := user.Current()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get current user: %w", err)
+		}
+
+		macOSSocketPath := filepath.Join(user.HomeDir, "Library/Containers/com.docker.docker/Data/docker.raw.sock")
```
willmurphyscode, line 71: `user.Homedir` has some edge cases that have bitten me on previous projects. I believe we used https://github.com/mitchellh/go-homedir for a fix.

internal/docker/docker_client_test.go
```diff
@@ -0,0 +1,44 @@
...
+				t.Errorf("Error: %v", err)
+			}
+			if dockerClient == nil {
+				t.Errorf("Error: dockerClient is nil")
+			}
+			daemonHost := dockerClient.DaemonHost()
+			if daemonHost == "" {
+				t.Errorf("Error: daemonHost is empty")
+			}
+			if strings.Contains(daemonHost, c.expectedSocket) == false {
```
willmurphyscode, line 39: Can we use a more specific assertion, like ends with?

internal/docker/docker_client_test.go
```diff
@@ -0,0 +1,44 @@
...
+	}{
+		{
+			name:           "Test newClient with default options",
+			osCase:         "",
+			expectedSocket: "unix:///var/run/docker.sock",
+		},
+		{
+			name:           "Test newClient with runtime specific path",
+			osCase:         "darwin",
+			expectedSocket: "Library/Containers/com.docker.docker/Data/docker.raw.soc",
```
willmurphyscode, line 22: I think this is missing a `k` on the end.

internal/docker/client.go
```diff
@@ -48,10 +51,45 @@ func GetClient() (*client.Client, error) {
...
+		if err == nil {
+			return dockerClient, nil // Successfully connected
+		}
+	}
+
+	// If both attempts failed
+	return nil, fmt.Errorf("failed to connect to Docker daemon. Ensure Docker is running and accessible")
+}
+
+func newClient(os string, opts ...client.Opt) (*client.Client, error) {
```
willmurphyscode, line 78: It surprised me that `os` changes between calls during the same invocation. Can we find a way to refactor so that this parameter never lies? (Passing `""` as `os` when we know perfectly well there is a value in runtime.GOOS feels like lying to the method.)

My suggestion to separate like this:
1. Get a []string of possible socket paths. This method has os specific behavior, and maybe accepts an `os` string parameter so it's easier to unit test.
2. Loop over them until one works. This method doesn't have os specific behavior.

internal/docker/client.go
```diff
@@ -48,10 +50,55 @@ func GetClient() (*client.Client, error) {
...
 	if err != nil {
-		return nil, fmt.Errorf("failed create docker client: %w", err)
+		return fmt.Errorf("failed to ping Docker daemon: %w", err)
 	}
+	return nil
+}
 
-	return dockerClient, nil
+/*
+Things I want to try
```
wagoodman, line 78: dead(ish) comment?

internal/docker/docker_client_test.go
```diff
@@ -0,0 +1,91 @@
...
+			expectedSocket: "unix:///var/run/docker.sock",
+		},
+		{
+			name:           "Test newClient with runtime specific path",
+			providedSocket: "",
+			setEnv: func() func() {
+				os.Setenv("DOCKER_HOST", "unix:///var/CUSTOM/docker.sock")
+				return func() {
+					os.Unsetenv("DOCKER_HOST")
+				}
```
wagoodman, line 32:
```suggestion
			setEnv: func(t *testing.T) func() {
				t.Setenv("DOCKER_HOST", "unix:///var/CUSTOM/docker.sock")
```

- testing will automatically re-set the value back to what it was before the test
- pass `t *testing.T` explicitly so that it is scoped to the right test at cleanup

internal/docker/docker_client_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/docker/docker/client"
+)
+
+// Write a test that asserts that we are not overwriting the DOCKER_HOST environment variable
```
wagoodman, line 11: dead comment?

wagoodman (approved):

internal/docker/client.go
```diff
@@ -48,10 +50,50 @@ func GetClient() (*client.Client, error) {
...
+	opts = append(opts, client.WithHost(socket))
+	return client.NewClientWithOpts(opts...)
+}
+
+func possibleSocketPaths(os string) []string {
+	switch os {
+	case "darwin":
+		hDir, err := homedir.Dir()
+		if err != nil {
+			return nil
```
willmurphyscode, line 90: Should this return `[]string{""}` instead? Right now an error fetching the home directory will prevent the attempt to use the default/fromEnv socket on macOS.

PR #219, opened 2024-02-06
feat: separate tagged image providers

kzantow:
This PR makes the following top-level changes:

* `image.Provider` interface has been reworked so providers have `Name()` method
* `image.Provider.Provide` is required to return an image object fully populated (e.g. `.Read()` has been called)
* A top-level `stereoscope.ImageProviders` function is added, allowing interested parties (e.g. syft) to get access to the individual tagged image providers
* A top-level `stereoscope.ImageProviderConfig` is added which is the uber-configuration needed by all stereoscope-defined providers

Note: the top-level `GetImage` and `GetImageFromSource` remain unchanged with the caveat that the `image.Source` has changed from an `int` to a `string`. These functions will use the default set of image providers, so should continue to behave about the same as before, with the other caveat that if the user has a docker daemon _and_ another supported daemon (containerd, podman), each daemon may be tried instead of stopping after attempting docker, which is the behavior today.

TODO:
- [x] ~~should `image.Provider` have a `Tags() []string` method and `ImageProviders(...)` returns `[]image.Provider`?~~ (no, this is now `[]collections.TaggedValue[image.Provider]`)
- [x] fix existing tests
- [x] add more tests exercising the full functionality of each image provider
- [x] fully integrate with syft PR https://github.com/anchore/syft/pull/2610


tagged/tagged.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+	}
+	return false
+}
+
+// HasValue indicates any of the provided values is present in this set
+func (t Values[T]) HasValue(value ...T) bool {
+	for _, tagged := range t {
+		for _, v := range value {
+			if isEqual(tagged.Value, v) {
```
kzantow, line 60: since `T` is `comparable`, `tagged.Value == v` works here, but unfortunately may still result in a panic by the go runtime in certain cases, so I've left a function to compare items if this needs to get worked around, it might be better to just make this use `==`

providers.go
```diff
@@ -0,0 +1,50 @@
...
+	RegistryTag = "registry"
+)
+
+// ImageProviderConfig is the uber-configuration containing all configuration needed by stereoscope image providers
+type ImageProviderConfig struct {
+	Registry image.RegistryOptions
+	Platform *image.Platform
+}
+
+func ImageProviders(cfg ImageProviderConfig) tagged.Values[image.Provider] {
```
kzantow, line 28: Should this signature return `(tagged.Values[image.Provider], error)`? this makes usage less convenient and we do not actually return an error here in any of the code paths at present.

wagoodman:
> should image.Platform be a parameter to the Provide function instead of configuration?

yeah, I think so

wagoodman:
> naming: is Resolver better than Provider (or something else)?

I'm a fan of "provider" here, but will give it some thoughts

tagged/tagged.go
```diff
@@ -0,0 +1,131 @@
+package tagged
+
+import (
+	"fmt"
+	"reflect"
+)
+
+// Value holds an arbitrary value with associated tags
+type Value[T comparable] struct {
```
wagoodman, line 9: I feel this should be broken out into a repo separate from stereoscope if we're going to use it in a more generic (ha) sense

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/anchore/stereoscope
 
-go 1.19
+go 1.20
```
wagoodman, line 3: there are some trixy generics bugs in 1.20 that I'm not certain were back ported, we might want to set this as 1.21 (don't consider this as a must-do comment, but a consideration that we need to validate)

client.go
```diff
@@ -69,151 +63,50 @@ func WithPlatform(platform string) Option {
 	}
 }
 
+// GetImage parses the user provided image string and provides an image object;
+// note: the source where the image should be referenced from is automatically inferred.
+func GetImage(ctx context.Context, imgStr string, options ...Option) (*image.Image, error) {
+	return GetImageFromSource(ctx, imgStr, "", options...)
```
wagoodman, line 69: I see it's convenient to allow for an empty string to be passed and it stands in for 'any', but given the function is called `GetImageFromSource` and should be distinctly different from `GetImage` in this particular functionality, I think a third unexported function should really be created that does the common functionality and the empty source checks end up being an explicit guard in the `GetImageFromSource` function.

.golangci.yaml
```diff
@@ -1,15 +1,17 @@
-run:
-  timeout: 10m
-linters-settings:
-  funlen:
-    lines: 75
+issues:
```
kzantow, line 1: FWIW this file was coped from [what's in Syft](https://github.com/anchore/syft/blob/main/.golangci.yaml)

providers.go
kzantow: No, I tried this and it makes handling errors much more confusing and defining a set of providers much more confusing. This effectively is only the glue code to provide necessary configuration to downstream functions.

tagged/tagged.go
```diff
@@ -0,0 +1,144 @@
...
+		Value: value,
+		Tags:  tags,
+	}
+}
+
+// ValueSet is a utility to handle a sorted set of tagged items including basic filtering
+type ValueSet[T comparable] []Value[T]
+
+// Collect returns a slice containing the values in the set
+func (t ValueSet[T]) Collect() []T {
```
wagoodman, line 39: semantically `Values()` makes a lot more sense here

pkg/image/image.go
```diff
@@ -159,6 +159,19 @@ func New(image v1.Image, tmpDirGen *file.TempDirGenerator, contentCacheDir strin
 	return imgObj
 }
 
+func Read(image *Image) (*Image, error) {
```
wagoodman, line 162: why create a static method to expose a public method on image when you already have the image?

kzantow: Removed, I just missed this.

client.go
```diff
@@ -51,13 +46,6 @@ func WithCredentials(credentials ...image.RegistryCredentials) Option {
 	}
 }
 
-func WithAdditionalMetadata(metadata ...image.AdditionalMetadata) Option {
```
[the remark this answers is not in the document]
kzantow: Restored this 👍

wagoodman (approved):

Issue #221, opened 2024-02-13
Add release process


wagoodman:
Currently stereoscope is unversioned (with the exception of the first tag manually performed). Moving forward we will be committing to tagged releases of stereoscope periodically. This issue represents the work to add the workflows, process docs, and helper scripts to facilitate releases.

PR #224, opened 2024-02-16
Upgrade tool management

wagoodman:
Upgrade practices similar to that of syft and grype, specifically:
- use binny to manage local tooling
- use task over make for coordinating tasks

This helps setup getting a release pipeline into stereoscope #221 , since it will be quite similar to how syft and grype release workflow-wise, getting the practices and tooling up to date first makes the transition easier.


.github/actions/bootstrap/action.yaml
```diff
@@ -1,80 +1,60 @@
...
+
 description: "Bootstrap all tools and dependencies"
 inputs:
   go-version:
     description: "Go version to install"
     required: true
-    default: "1.19.x"
-  use-go-cache:
-    description: "Restore go cache"
+    default: "1.21.x"
```
spiffcs, line 8: Is 1.22 available yet?

Makefile
```diff
@@ -1,208 +1,46 @@
-TEMP_DIR = ./.tmp
+OWNER = anchore
+PROJECT = syft
```
spiffcs, line 2: I think this needs to be `stereoscope`

Taskfile.yaml
```diff
@@ -0,0 +1,311 @@
...
+
+  # note: the snapshot dir must be a relative path starting with ./
+  SNAPSHOT_DIR: ./snapshot
+  SNAPSHOT_CMD: "{{ .TOOL_DIR }}/goreleaser release --config {{ .TMP_DIR }}/goreleaser.yaml --clean --snapshot --skip=publish --skip=sign"
+  RELEASE_CMD:  "{{ .TOOL_DIR }}/goreleaser release --clean --release-notes {{ .CHANGELOG }}"
+  VERSION:
+    sh: git describe --dirty --always --tags
+
+  # used for install and acceptance testing
+  COMPARE_DIR: ./test/compare
```
spiffcs, line 23: Are these used by stereoscope at all?

spiffcs (commented):
small comments added with questions about use of variables and potential project name conflict

Makefile
wagoodman: nice find

Taskfile.yaml
wagoodman: no,I'll remove these... I'll do one more pass for syft-specific things that got copied in

spiffcs (approved):
👍 with comments addressed

PR #228, opened 2024-04-04
Fix panic when pulling OCI-packaged helm chart

willmurphyscode:
Reported in https://github.com/anchore/syft/issues/2745

## TODO

- [ ] add some unit tests
- [ ] get feedback on Syft e2e behavior when using this change (currently is sort of a strange multierror output that seems hard to parse for me):

``` sh
❯ go run cmd/syft/main.go bitnamicharts/nginx:15.14.1
 ✔ Pulled image                    
 ✔ Parsed image                                                                                                                   sha256:fa265257e1b905d79242f2b25b506057d179a108831b99f27a2085fc02706fff
could not determine source: errors occurred attempting to resolve 'bitnamicharts/nginx:15.14.1':
  - no such file or directory
  - unable to inspect existing image: Error response from daemon: No such image: bitnamicharts/nginx:15.14.1
  - podman not available: making http client: connection to bastion host="127.0.0.1:57447" failed: dial tcp 127.0.0.1:57447: connect: connection refused
  - containerd not available: no grpc connection or services is available: unavailable
  - unknown layer media type: application/vnd.cncf.helm.chart.content.v1.tar+gzip
exit status 1
```

IMO this is sort of confusing output. The last line "unknown layer media type: application/vnd.cncf.helm.chart.content.v1.tar+gzip" seems the most informative.

Basically, this is the error because stereoscope tried to use all its image providers to pull the image, and they all failed with various things, and Syft prints them all because it doesn't know which one is the "real" error. Probably more work is needed to translate this to a higher level error message in Syft. For example, the error has nothing to do with podman not being installed, but that's reported with the same UI prominence as the unsupported media type error.


willmurphyscode:
The new syft error handling (PR soon) will look like this:

```
❯ go run cmd/syft/main.go bitnamicharts/nginx:15.14.1
 ✔ Pulled image                    
 ✔ Parsed image                                                                                                                                                                                                                                                                                                                                                                                           sha256:fa265257e1b905d79242f2b25b506057d179a108831b99f27a2085fc02706fff
could not determine source: errors occurred attempting to resolve 'bitnamicharts/nginx:15.14.1':
  - docker: unable to inspect existing image: Error response from daemon: No such image: bitnamicharts/nginx:15.14.1
  - podman: podman not available: making http client: connection to bastion host="127.0.0.1:57447" failed: dial tcp 127.0.0.1:57447: connect: connection refused
  - containerd: containerd not available: no grpc connection or services is available: unavailable
  - oci-registry: unknown layer media type: application/vnd.cncf.helm.chart.content.v1.tar+gzip
  - additionally, the following providers failed with file does not exist: docker-archive, oci-archive, oci-dir, singularity, oci-dir, local-file, local-directory
```

Because we don't know which provider the user was expecting to handle the source, report each error with the name of the provider that caused it, except for providers that fail with a file not found error - just report a list of those, since every provider that assumes the input string was a path on the local filesystem will fail with that error if it wasn't.

pkg/image/layer.go
```diff
@@ -107,44 +99,82 @@ func (l *Layer) Read(catalog *FileCatalog, imgMetadata Metadata, idx int, uncomp
...
+	)
+	if err != nil {
+		return fmt.Errorf("failed to read layer=%q tar : %w", l.Metadata.Digest, err)
+	}
+
 	monitor.SetCompleted()
+	return nil
+}
+
+func (l *Layer) readSingularityImage(idx int, tree *filetree.FileTree) error {
```
kzantow, line 149: I see that some of this is a refactoring of what's already there, but we should not be provider-specific code in the main `image` package. We should probably just get rid of the `Read()` call and move appropriate "reading" into the appropriate image providers. Would that help to return more appropriate errors?

pkg/image/layer_metadata.go
```diff
@@ -16,17 +16,19 @@ type LayerMetadata struct {
...
+	if err != nil {
+		return LayerMetadata{}, err
+	}
 
-	// digest = diff-id = a digest of the uncompressed layer content
-	diffIDHash := imgMetadata.Config.RootFS.DiffIDs[idx]
 	return LayerMetadata{
 		Index:     uint(idx),
-		Digest:    diffIDHash.String(),
+		Digest:    diffID.String(),
```
kzantow, line 31: I believe the most important part of this change is here, where there was a panic on line 26. Are we sure that in all cases the `layer.DiffID()` is going to return the same value (e.g. docker-archive, oci-archive, oci-directory, etc.)? I really don't know the right answer, but would be really wary about changing this logic. Could this PR just be simplified a bit to maybe check the `imgMetadata` and sub-properties and return an error in the event they are nil? Or _then_ try the `layer.DiffID`?

willmurphyscode: Thanks for the feedback here @kzantow 

1. Do you know why stereoscope's code is reading the diff ID from the image metadata, rather than asking the layer for the diff ID? The panic seems like evidence that we're handling this incorrectly.
2. If indexing reading into the rootFS diff IDs returns a different answer than layer.DiffID, isn't the layer right?

I can revert this, and maybe I should. The panic will actually be fixed in this PR by checking the media type before calling `newLayerMetadata`.

The panic was because there is no RootFS for helm charts packed as OCI images, AFAICT. Calling layer.DiffID() on one returns an error.

willmurphyscode: @kzantow So I did the following experiment:

In my local code, I added a function like this:

``` go
func panicOnDiffIDMismatch(l *Layer, imgMetadata Metadata, idx int) {
	interfaceDiffID, err := l.layer.DiffID()
	if err != nil {
		panic(err)
	}

	homebrewDiffID := imgMetadata.Config.RootFS.DiffIDs[idx]

	if homebrewDiffID != interfaceDiffID {
		panic("everything we know is wrong!")
	}
}
```

and called it at the top of `layer.Read`.

With this change included, Syft and Stereoscope integration tests pass, as long as I set `CONTAINER_HOST` to the socket podman is actually listening on, which is a requirement on `main` as well AFAICT.

I think I'm in favor of changing our implementation to use `layer.DiffID()` instead of our own diff ID:

1. If the two methods are always the same, it doesn't matter
2. If the two methods _aren't_ always the same, isn't the one implemented in `go-containerregistry` more likely correct?

On the other hand, I understand your caution about changing this. What do you think?

wagoodman: agreed to use `layer.DiffID`, the original reasons for not doing so are lost to history (though it might have been for performance, but I'm not seeing anything in a code review in GGCR for this to be true).

pkg/image/layer.go
wagoodman: this does make some sense, but I'm not certain it should be done on this PR (this suggestion needs to be more specific on what the impacts are, which isn't really due for this fix)

PR #234, opened 2024-06-18
Bump docker version

jonjohnsonjr:
Ran into an issue compiling a binary that includes anchore because our version of docker no longer compiles since types.ImagePullOptions got moved to a different package. This bumps docker and updates to use the newer package options.

I had to update the singularity thing due to this error:

> While performing build: conveyor failed to get: initializing source docker-daemon:stereoscope-fixture-image-symlinks:b6d810e8a6f2baa497a6b5429c65467b2df56f13558247ad2ed23f9133186a6f: loading image from docker engine: Error response from daemon: {"message":"client version 1.22 is too old. Minimum supported API version is 1.24, please upgrade your client to a newer version"}


wagoodman (approved):
Thanks for the version bump 🙌

PR #258, opened 2024-07-04
update docker deps from v26 to v27

k4leung4:


kzantow:
Thanks for the PR @k4leung4. It looks like dependabot got these updated already, so I'm going to close this PR, let me know if I've missed something!

PR #264, opened 2024-07-08
chore: grouped dependabot updates

kzantow:
This PR enables [grouped dependabot updates](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,7 +12,6 @@ updates:
     groups:
       go:
         update-types:
-          - major
           - minor
           - patch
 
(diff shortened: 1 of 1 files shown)


wagoodman (commented):
I'm down for grouping dep PRs -- should we group minor and patch into a single PR and keep major versions in a separate PR (since those have the likely hood of breaking things per semver)?

kzantow:
@wagoodman I'm ok making more changes here, but for the time being I updated the `go` updates to exclude grouping the major version bumps but left this for action updates. Looking at our [current pull requests](https://github.com/anchore/stereoscope/pulls), there are multiple major action bumps which are probably not causing problems, but 3 go updates to major and minor versions which do not pass all checks. Happy to make more changes here, just trying to reduce the noise as much as possible.

wagoodman (approved):
Thanks for thinking of this... there are a lot of dependabot PRs 😳

PR #265, opened 2024-07-08
deps: bump github.com/sylabs/squashfs to `v1.0.0`

tri-adam:
#261 pulls in an updated `github.com/sylabs/squashfs ` dependency, but unfortunately the v1.0.0 of that has some breaking API changes! I've attempted to address those here, and made some other improvements to the code that backs the `singularity` source. In particular:

- I've added an `error` to the `file.Opener`, to be able to explicitly pass an error back when encountered reading a SquashFS file. Let me know if this looks OK... I had to make minor mods in other sources to satisfy that interface, but seems more consistent with how things like `go-containerregistry` return `io.ReadCloser`s.
- I've generalized what was `uncompressedTarCache` to be `uncompressedCache`, and used that on the SquashFS side of things to make sure the code closes file handles everywhere it should.

One last note on testing, not sure the best way to do this? I built `syft` with this version like so:

```
$ go mod edit -replace github.com/anchore/stereoscope=../stereoscope
```

And then pulled a SIF image and scanned it:

```
$ singularity pull docker://alpine
...
$ go run ./cmd/syft scan singularity:alpine_latest.sif
 ✔ Parsed image                                                                                                                                                   sha256:bd640a41460ae7bfae0d608ad063c84cf5b61ef1b9647531bd59cbf57e519dde
 ✔ Cataloged contents                                                                                                                                                    ff77047d9cb035df6be6800b79aea4df63d2fb3e46c59125b2fc2d5235062496
   ├── ✔ Packages                        [14 packages]  
   ├── ✔ File digests                    [77 files]  
   ├── ✔ File metadata                   [77 locations]  
   └── ✔ Executables                     [17 executables]  
NAME                    VERSION      TYPE   
alpine-baselayout       3.6.5-r0     apk     
alpine-baselayout-data  3.6.5-r0     apk     
alpine-keys             2.4-r1       apk     
apk-tools               2.14.4-r0    apk     
busybox                 1.36.1-r29   apk     
busybox-binsh           1.36.1-r29   apk     
ca-certificates-bundle  20240226-r0  apk     
libcrypto3              3.3.1-r0     apk     
libssl3                 3.3.1-r0     apk     
musl                    1.2.5-r0     apk     
musl-utils              1.2.5-r0     apk     
scanelf                 1.3.7-r2     apk     
ssl_client              1.36.1-r29   apk     
zlib                    1.3.1-r1     apk
```

Let me know if that looks alright, or if there are any other improvements I can make. Thanks!


wagoodman:
Thanks for making the bump @tri-adam !

> One last note on testing, not sure the best way to do this? I built syft with this version like so:
> $ go mod edit -replace github.com/anchore/stereoscope=../stereoscope

indeed, that way works well for local testing. Another way is to use [`go.work`](https://go.dev/doc/tutorial/workspaces):
```
# from the syft dir...
go work init

# use "current" dir and "../stereoscope" within the workspace
go work use . ../stereoscope
```

This has the added benefit of not changing the `go.mod` and `go.sum` files directly, so is a little easier for local development.

PR #268, opened 2024-07-10
Shallow clone trees to save on memory usage

### LINKED ISSUE #233: Very High Memory Usage Using Syft
**What happened**:
High memory consumption when scanning 5GB image

controlled memory consumption

**Steps to reproduce the issue**:
scan very big sized images and check the memory consumption

image

**Environment**:
- Output of `syft version`: 1.5.0
- OS (e.g: `cat /etc/os-release` or similar):
seems like there are many dotnet packages in the images

wagoodman:
Based off of the inuse memory in stereoscope we spend of lot of memory on copying the filetree for squashing. An observation here is that all tree mutation operations always create or replace nodes, but never mutate existing nodes. This means that we can share references across the multiple trees created for the squashfs functionality for each layer.

Here is a snapshot of inuse memory today for image read for a sample image:
Screenshot 2024-07-10 at 1 48 13 PM

Here is a snapshot of inuse memory after using `map.Clone` to shallow clone existing trees:
Screenshot 2024-07-10 at 1 48 22 PM

Partially fixes #233 in terms of inuse memory

### DIFF
--- a/pkg/filetree/filenode/filenode.go
+++ b/pkg/filetree/filenode/filenode.go
@@ -55,15 +55,6 @@ func (n *FileNode) ID() node.ID {
 	return IDByPath(n.RealPath)
 }
 
-func (n *FileNode) Copy() node.Node {
-	return &FileNode{
-		RealPath:  n.RealPath,
-		FileType:  n.FileType,
-		LinkPath:  n.LinkPath,
-		Reference: n.Reference,
-	}
-}
-
 func (n *FileNode) IsLink() bool {
 	return n.FileType == file.TypeHardLink || n.FileType == file.TypeSymLink
 }
--- a/pkg/filetree/filenode/filenode.go
+++ b/pkg/filetree/filenode/filenode.go
@@ -55,6 +55,15 @@ func (n *FileNode) ID() node.ID {
 	return IDByPath(n.RealPath)
 }
 
+func (n *FileNode) Copy() node.Node {
+	return &FileNode{
+		RealPath:  n.RealPath,
+		FileType:  n.FileType,
+		LinkPath:  n.LinkPath,
+		Reference: n.Reference,
+	}
+}
+
 func (n *FileNode) IsLink() bool {
 	return n.FileType == file.TypeHardLink || n.FileType == file.TypeSymLink
 }
--- a/pkg/filetree/filetree.go
+++ b/pkg/filetree/filetree.go
@@ -45,10 +45,10 @@ func New() *FileTree {
 	}
 }
 
-// Copy returns a Copy of the current FileTree.
-func (t *FileTree) Copy() (ReadWriter, error) {
+// Clone returns a shallow Copy of the current FileTree.
+func (t *FileTree) Clone() (ReadWriter, error) {
 	ct := New()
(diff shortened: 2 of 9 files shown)


kzantow (approved):

wagoodman:
After talking with @willmurphyscode I'm going to hold off on merging in order to demonstrate a little more proof of this working as intended in syft. If all goes well I think this will land on the order of days 🎉

willmurphyscode:
The test we still want to do is get a big image that has different file types for the same location, and edit the file, and prove that the filetree object doesn't get mutated. This can probably be done with the [stereoscope example application](https://github.com/anchore/stereoscope/blob/main/examples/basic.go). Basically, make an image that, for a given path, has layers where that path is a directory with children, then where it's a file, where it's a link, where it's a file with different contents, etc. Basically fuzzing the tree to prove that the structs that are now being shallow-copied won't be mutated.

willmurphyscode:
I ran a test, and I think we have some differences here.

## Overview

1. build an image that does silly things with a path (e.g. there's a path that's a symlink in one layer, an empty dir in another, a dir with contents in another, a file in another). Dockerfile below.
2. Check out `main` and write `go run ./examples/basic.go` and save the result
3. Check out this PR, rebase on `main`, and then run `go run ./examples/basic.go` and save the result
4. Diff the files from steps 2 and 3. If there's no diff, the change has no effect.

## Setup

``` sh
❯ tree .
.
├── Dockerfile
└── chimera
    ├── a.txt
    └── b.txt
```

``` Dockerfile
FROM debian:trixie-slim

# makes /chimera/a.txt /chimera/b.txt
COPY . / 

# make a layer where the dir is empty
RUN rm /chimera/a.txt /chimera/b.txt

RUN rmdir /chimera

RUN touch /chimera

RUN rm /chimera

RUN ln -s /etc/hostname /chimera

RUN unlink /chimera

RUN mkdir /chimera
```

```
docker build . -t localhost/aggressive
```

Saving results like

``` sh
go run ./examples/basic.go localhost/aggressive > willtmp/before.txt
git checkout shallow-tree-clone
git rebase origin/main
go run ./examples/basic.go localhost/aggressive > willtmp/after-with-rebase.txt
```

## Results

Running `diff -u --color willtmp/before.txt willtmp/after-with-rebase.txt` produces the following output:

``` diff
--- willtmp/before.txt  2024-11-23 07:39:21
+++ willtmp/after-with-rebase.txt       2024-11-23 07:44:20
@@ -10336,8 +10336,6 @@
     /bin/znew
     /boot
     /chimera
-    /chimera/a.txt
-    /chimera/b.txt
     /dev
     /etc
     /etc/.pwd.lock
```

I think some more investigation is needed before we merge this. A reasonable next step is probably to write a failing test that exhibits this bug and push it to this branch. The above testing proves there is a difference, but not much about what the cause is.

willmurphyscode:
I've pushed a test case that passes on `main` but fails on this branch that I think illustrates the bug.

wagoodman:
Thanks for taking the due diligence here and proving that this doesn't quite work in the context of syft -- I'm going to close this for now.

PR #314, opened 2024-10-27
allow configuring larger max layer size in build time

### LINKED ISSUE #311: increase max layer/file size from 2GB
Support for layers above 2GB

**Why is this needed**:
Syft fails for such images

**Additional context**:
It happens due to [hard-coded](https://github.com/anchore/stereoscope/blob/main/pkg/file/tarutil.go#L17) threshold.
Apparently in [syft ](https://github.com/anchore/syft/blob/main/internal/file/copy.go#L9) there is also a limit for tar/zip files size, but I think it's less likely to happen.

One of the options is to make this limit configurable or at least to increase it e.g. to 20GB.

2rigor:
Closes #311

E.g., for 10Gb (10*10^9 bytes) append to go build command:
`-ldflags "-w -s -X github.com/anchore/stereoscope/pkg/file.perFileReadLimitStr=10000000000"`


2rigor:
@wagoodman I don't know how it works. I see you were involved in many PRs. Will you guys automatically have a look at new PR or I should assign it (to whom)? Thanks!

kzantow (changes requested):
Hey @2rigor -- thanks for the contribution. We might like to take a different approach to allow for modifying the per-file read limit. Specifying this at build time doesn't allow configuration at runtime, which is probably the most important spot.

What if instead of making the change as you've implemented it, we just change the `perFileReadLimit` to a `var`, and add an exported function to set it; e.g.:
```
var perFileReadLimit = 2 * GB

func SetPerFileReadLimit(maxBytes int64) {
  // some reasonable validation like maxBytes > 0?
  perFileReadLimit = maxBytes
}
```

... this allows library users including Syft to configure this value via configuration files at runtime.

kzantow:
> but having such a function provides an option to change the variable at any time. I still think it's acceptable. Just wanted to share my thoughts.

Yes, I don't think this is a problem, though -- if the value gets changed, it gets changed for the next reader; I don't think there's a concurrency issue with this. You can set this at initialization time by creating an `init()` function in your code.

2rigor:
> > but having such a function provides an option to change the variable at any time. I still think it's acceptable. Just wanted to share my thoughts.
> 
> Yes, I don't think this is a problem, though -- if the value gets changed, it gets changed for the next reader; I don't think there's a concurrency issue with this. You can set this at initialization time by creating an `init()` function in your code.

Got you, I agree it's better. Implemented as you originally suggested

kzantow (approved):
LGTM -- thanks for the contribution @2rigor !

tomersein:
how can i set this value via syft? @kzantow

kzantow:
@tomersein -- there is no way to configure this in Syft at the moment, but you could add a configuration to [one of the options](https://github.com/anchore/syft/tree/main/cmd/syft/internal/options) that in a `PostLoad()` hook, reads a value and sets this using the stereoscope hook that was added.

tomersein:
hi, the 1st part I think I understand, maybe under catalogers field (because it is related to scan)
but didn't understand the 2nd part, where is the hook of stereoscope and how can i add it? :) @kzantow

PR #336, opened 2025-01-04
Add more strict verifications when user provides a platform

wagoodman:
Today we allow for users to specify a platform for images being resolved. The expectation is that if the fetched/resolved image does not match the requested platform then we should error out and not continue. Bugs were found in both how the Docker daemon provider (thus Podman as well since this is shared code) as well as the OCI registry provider. Though the bugs were slightly different, at a high level they were the same: a user could provide a platform to fetch and the provider in some cases would ignore this platform and fetch the image for a different platform.

For the Docker daemon provider when monitoring the pull event status JSONL stream the `Error` field was not being considered, which is where this kind of error is raised up.

For the OCI registry provider, passing the `remote.WithPlatform` option only sets the platform field on the `v1.Descriptor` object, but does not have any effect on what is fetched in the case where there is no manifest list or index (there is a single architecture in the registry). In the case of manifest list or index the correct platform was being resolved. The change here was to explicitly pull the container config and validate against the `os` and `architecture` fields (which are required via the OCI spec).

An additional step within the OCI registry provider is checking if the manifest is a list/index or a single manifest. If the user does NOT give a platform then the current default behavior for this provider is to set one based off of `linux/<GOARCH>`. The intended behavior is to honor what the single-architecture manifest instead of overriding with a default value. This PR enforces this intended behavior by clearing the platform options after fetching the manifest and checking the MediaType for single vs multi arch support.

~The last change is to allow for providers to express errors that indicate that the image was fully resolved by the provider but there was a transient or fundamental error. This new error is intended to signal to the caller that no further providers in a set of providers should be attempted, and to raise up this single error. For instance, if we tried 7 providers with 7 "not found" errors and we reach the Docker provider and it raises up a platform mismatch error, we don't want to attempt the Registry provider since we expect a similar outcome. This fosters a fail-fast approach for cases we know the user should be paying attention to (and not get lost in a pile of errors from multiple providers).~ Instead a new specific error type has been added to allow the caller to more clearly reason about platform mismatch errors.

### DIFF
--- a/pkg/image/provider.go
+++ b/pkg/image/provider.go
@@ -2,7 +2,21 @@ package image
 
 import (
 	"context"
+	"fmt"
 )
+
+// ErrFetchingImage is meant to be used when a provider has positively resolved the image but, while fetching the
+// image, an error occurred. The goal is to differentiate between a provider that cannot resolve an image (thus
+// if the caller has a set of providers, it can try another provider) and a provider that can resolve an image but
+// there is an unresolvable problem (e.g. network error, mismatched architecture, etc... thus the caller should
+// not try any further providers).
+type ErrFetchingImage struct {
+	Reason string
+}
+
+func (e *ErrFetchingImage) Error() string {
+	return fmt.Sprintf("error fetching image: %s", e.Reason)
+}
 
 // Provider is an abstraction for any object that provides image objects (e.g. the docker daemon API, a tar file of
 // an OCI image, podman varlink API, etc.).
--- a/pkg/image/provider.go
+++ b/pkg/image/provider.go
@@ -5,17 +5,22 @@ import (
 	"fmt"
 )
 
-// ErrFetchingImage is meant to be used when a provider has positively resolved the image but, while fetching the
-// image, an error occurred. The goal is to differentiate between a provider that cannot resolve an image (thus
-// if the caller has a set of providers, it can try another provider) and a provider that can resolve an image but
-// there is an unresolvable problem (e.g. network error, mismatched architecture, etc... thus the caller should
-// not try any further providers).
-type ErrFetchingImage struct {
-	Reason string
+// ErrPlatformMismatch is meant to be used when a provider has positively resolved the image but the image OS or
+// architecture does not match with what was requested.
+type ErrPlatformMismatch struct {
+	ExpectedPlatform string
+	Err              error
 }
 
-func (e *ErrFetchingImage) Error() string {
-	return fmt.Sprintf("error fetching image: %s", e.Reason)
+func (e *ErrPlatformMismatch) Error() string {
+	if e.ExpectedPlatform == "" {
+		return fmt.Sprintf("mismatched platform: %v", e.Err)
+	}
+	return fmt.Sprintf("mismatched platform (expected %v): %v", e.ExpectedPlatform, e.Err)
+}
+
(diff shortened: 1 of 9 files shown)


pkg/image/provider.go
```diff
@@ -2,8 +2,22 @@ package image
...
 	"context"
+	"fmt"
 )
 
+// ErrFetchingImage is meant to be used when a provider has positively resolved the image but, while fetching the
+// image, an error occurred. The goal is to differentiate between a provider that cannot resolve an image (thus
+// if the caller has a set of providers, it can try another provider) and a provider that can resolve an image but
+// there is an unresolvable problem (e.g. network error, mismatched architecture, etc... thus the caller should
+// not try any further providers).
+type ErrFetchingImage struct {
```
kzantow, line 13: I'm not sure we should add behavior that could result in one provider causing other providers to get bypassed. The source search, to me, is "try whatever you can to find the right thing". Part of the reason to isolate providers to the specific thing they provide is to make it easier to add other providers that might fill in gaps these providers have and isolate the logic so they shouldn't affect other providers.

The scenario, as I understand it is this: we attempt to resolve an image, but we can't because we can't match the correct platform. So, say the docker daemon provider determines this, as an example... what if we built a matching image _without_ the correct platform locally. So running `syft <image>` would resolve our locally built image, but fails because the platform was wrong. If we return a "stop" error, the registry provider doesn't run and we potentially don't resolve a valid image that existed in a remote registry. I don't think we should assume the user wanted the locally built image -- they can easily restrict what they're searching using `--from docker` or `--from registry` to force specific providers if there's a reason to, like only checking a locally-built image.

I _think_ the thing this is trying to prevent is _downloading_ large files twice from two different providers, right? Is this because they don't know the platform until downloading the whole thing? This seems like a different change: the registry provider should download the container metadata to ensure it actually found the right thing, with the right platform, before downloading any layer blobs. Maybe this is a hard change, but is it really a problem if both of these download an image? To me this is eschewing performance in favor of correctness and I'd point out that I don't think it affects enterprise, since I don't think they are using the daemon provider at all, only syft users who are explicitly specifying an incorrect platform.

wagoodman: > If we return a "stop" error, the registry provider doesn't run and we potentially don't resolve a valid image that existed in a remote registry. 

That makes sense, I can back this part out -- but from a stereoscope point of view this isn't really changing behavior. It's allowing the caller to make this decision. We could leave this error in and not act on it in syft too as another option? But I agree with your overall point that within the stereoscope/syft ecosystem we're intending the providers to be independent of one another.

> the registry provider should download the container metadata to ensure it actually found the right thing, with the right platform, before downloading any layer blobs

actually, that is how it works today -- `remote.Get()` only grabs the manifest, then we now additionally grab the container config, but the layer blobs are not requested until we start indexing them (only the registry provider does this today, the others grab them all at once).

kzantow: Good point: the comment/behavior here isn't actually what would be responsible for it to happen; could this instead then capture the original error rather than just the message?

wagoodman: done!
```golang
// ErrPlatformMismatch is meant to be used when a provider has positively resolved the image but the image OS or
// architecture does not match with what was requested.
type ErrPlatformMismatch struct {
	ExpectedPlatform string
	Err              error
}
```

kzantow (approved):

PR #355, opened 2025-02-11
Remove logging for filetree search queries

wagoodman:
This PR removes some logging in common paths, which tends to lead to very verbose trace logs.


kzantow (approved):
Agree this is a bit excessive; removing is 👍

PR #379, opened 2025-03-12
fix: pull in sylabs squashs v1.0.5

willmurphyscode:
Pull in fix for infinite read bug in suqash fs reader implementation.

When pulled in to syft, will address anchore/syft#3651

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/anchore/stereoscope
 
-go 1.24.0
-
-toolchain go1.24.1
+go 1.24.1
 
 require (
 	github.com/GoogleCloudPlatform/docker-credential-gcr v2.0.5+incompatible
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/anchore/stereoscope
 
-go 1.22.9
+go 1.24.0
+
+toolchain go1.24.1
```
kzantow, line 5: FWIW I would rather avoid getting these `toolchain` entries, if possible. I thought that would happen with a third segment in the `go` directive like `go 1.24.0` 🤷

willmurphyscode: I'm not sure what I should do differently to avoid leaving that there - manually deleting it makes me feel like I'm missing something. I'll keep looking, but if you have steps for how to avoid this, please feel free to share :)

willmurphyscode: @kzantow making the `go` statement have `1.24.1` made the `toolchain` line go away.

kzantow (approved):
🎉

PR #381, opened 2025-03-13
Set the GGCR user agent.

mattmoor:
This is a feature that's generally encouraged for consumers of GGCR, so that registry vendors can better identify the clients that are using GGCR to talk to registries.

cc @luhring

### DIFF
--- a/pkg/image/oci/registry_provider.go
+++ b/pkg/image/oci/registry_provider.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
+	"os"
 	"runtime"
 
 	"github.com/google/go-containerregistry/pkg/authn"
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
Ideally you could include both `os.Args[0]` and the version, we do something similar for ggcr itself here: https://github.com/google/go-containerregistry/blob/098045d5e61ff426a61a0eecc19ad0c433cd35a9/pkg/v1/remote/transport/useragent.go#L23-L71

mattmoor:
Happy to include the version reflection logic if the maintainers want it, defer to them on preference.

wagoodman:
I think adding the application is a good start. If we're going to add version support that should probably be plumbed in via syft and grype directly (so expose a way to provide that via the API here).

PR #411, opened 2025-06-03
fix: complex symlink performance issue

kzantow:
This PR adjust how symlinks are matched against glob patterns such that we will now only return the first matching path for an index entry. This often requires no symlink resolution at all, for example when glob patterns match the real path. This solves a problem that is beginning to be apparent with complex webs of symlinks such as large pnpm projects and snaps that would result in an explosion of paths returned to be tested against a glob pattern causing excessive memory usage and processing time, which effectively would cause syft to hang.

### DIFF
--- a/pkg/filetree/search.go
+++ b/pkg/filetree/search.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path"
 	"sort"
+	"strings"
 
 	"github.com/bmatcuk/doublestar/v4"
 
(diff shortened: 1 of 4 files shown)


pkg/filetree/search.go
```diff
@@ -98,7 +99,7 @@ func (sc searchContext) SearchByMIMEType(mimeTypes ...string) ([]file.Resolution
 		fileEntries = append(fileEntries, entries...)
 	}
 
-	refs, err := sc.referencesInTree(fileEntries)
+	refs, err := sc.firstMatchingReferences("**/*", fileEntries)
```
kzantow, line 102: There was a problem with [the alternate change](https://github.com/anchore/stereoscope/commit/b9900fbe31293c779df37409e72bf09eee4b5b5d#diff-289f1bd7c87e8eec652c5d4db11c3eae9a3ed8d3a108e13dd16d2b18dd69ce67R102) -- it was returning too many results for different layers. I'm happy to figure that out, but this change also solves the performance issue, albeit with some unnecessary glob matching.

wagoodman: that's alright, the string glob match is fairly cheap when compared to other operations happening here

PR #417, opened 2025-06-17
Sh 3985 Daemon Cleanup Pulled Image feature flag (disabled by default)

houdini91:
Second Suggestion: Cleanup Only when active pull by provider.
See https://github.com/anchore/syft/issues/3985

### DIFF
--- a/pkg/image/containerd/daemon_provider.go
+++ b/pkg/image/containerd/daemon_provider.go
@@ -46,6 +46,7 @@ func NewDaemonProvider(tmpDirGen *file.TempDirGenerator, registryOptions image.R
 		platform:        platform,
 		namespace:       namespace,
 		registryOptions: registryOptions,
+		imageCleanup:    imageCleanup,
 	}
 }
 


pkg/image/containerd/daemon_provider.go
```diff
@@ -446,6 +450,16 @@ func (p *daemonImageProvider) saveImage(ctx context.Context, client *containerd.
 	return tempTarFile.Name(), nil
 }
 
+func (p *daemonImageProvider) deleteImage(ctx context.Context, client *containerd.Client, resolvedImage string) error {
+	// Remove the image
+	if err := client.ImageService().Delete(ctx, resolvedImage, nil); err != nil {
+		return fmt.Errorf("failed to delete image %s: %w", resolvedImage, err)
+	}
+
+	log.Infof("deleted image %s from containerd", resolvedImage)
```
kzantow, line 459: This provider is used for docker, podman, etc., so should use the `p.Name()` function instead of hardcoded `containerd`

kzantow:
Hey @houdini91 just noting here we will want to make sure this is disabled by default and opt in; I [noted in the original issue](https://github.com/anchore/syft/issues/3985#issuecomment-3009743785).

houdini91:
@kzantow I've updated the code to reflect your comment. If you prefer a different approach, please let me know. (One unexplored option in this PR was using an environment variable like TMPDIR, perhaps IMAGE_CLEANUP_ENABLE, to keep provider constructors unchanged.)"

dn-scribe:
A needed feature when running on CIs or when running to generate sboms at scale

houdini91:
Up

PR #431, opened 2025-07-28
fix: lazyBoundedReadCloser correctly close/seek

kzantow:
This PR fixes an issue where reading a lazyBoundedReadCloser is unable to seek due to being closed but not reset properly. In other words: `.Seek(0, io.SeekStart)` attempts to call `d.openFile()`, but since there is a non-nil reference to a reader, it does not reopen the file but fails to seek and read from it afterwards instead of actually reopening it for reading.

### DIFF
--- a/pkg/file/lazy_bounded_read_closer.go
+++ b/pkg/file/lazy_bounded_read_closer.go
@@ -4,11 +4,15 @@ import (
 	"errors"
 	"io"
 	"os"
+
+	"github.com/anchore/stereoscope/internal/log"
 )
 
-var _ io.ReadCloser = (*lazyBoundedReadCloser)(nil)
-var _ io.ReaderAt = (*lazyBoundedReadCloser)(nil)
-var _ io.Seeker = (*lazyBoundedReadCloser)(nil)
+var _ interface {
+	io.ReadCloser
+	io.ReaderAt
+	io.Seeker
+} = (*lazyBoundedReadCloser)(nil)
 
 // lazyBoundedReadCloser is a "lazy" read closer, allocating a file descriptor for the given path only upon the first Read() call.
 // Only part of the file is allowed to be read, starting at a given position.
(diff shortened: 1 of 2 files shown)


pkg/file/lazy_bounded_read_closer.go
```diff
@@ -80,10 +84,10 @@ func (d *lazyBoundedReadCloser) ReadAt(b []byte, off int64) (n int, err error) {
 
 	n, err = d.reader.ReadAt(b, off)
 	if err != nil && errors.Is(err, io.EOF) {
-		// we've reached the end of the file, force a release of the file descriptor. If the file has already been
-		// closed, ignore the error.
-		if closeErr := d.file.Close(); !errors.Is(closeErr, os.ErrClosed) {
```
spiffcs, line 85: Review note: `d.Close()` does not hoist from d.file.Close(). It's using: https://github.com/anchore/stereoscope/blob/main/pkg/file/lazy_bounded_read_closer.go#L53-L66

This function calls `d.file.Close()` while also resetting d.file and d.reader. 
`closeErr` cannot be `os.ErrClosed` given this is already accounted for by `d.Close()`

@kzantow what was the original issue where you saw this behave weird when it wasn't reset?

I just want to know how this manifested in the wild so I can understand the change before 🟢

pkg/file/lazy_bounded_read_closer.go
```diff
@@ -80,10 +84,10 @@ func (d *lazyBoundedReadCloser) ReadAt(b []byte, off int64) (n int, err error) {
...
-			return n, closeErr
+		// we've reached the end of the file, release of the file descriptor. continue to return EOF
+		// IMPORTANT: call d.Close to reset internal state
+		if closeErr := d.Close(); closeErr != nil {
+			log.Tracef("unable to close: %v: %v", d.path, closeErr)
```
kzantow, line 90: I'm not entirely sure the best behavior here, can certainly be convinced to change this... but this is only hit if we received EOF during the Read call, which will be returned to the caller and we're managing open file handles behind the scenes, so while an error closing here can indicate a problem, there's not much that a call can do about this and they are already receiving an EOF, so should stop further reading anyway unless they Seek, at which time we need to be able to reopen the file

spiffcs (approved):

pkg/file/lazy_bounded_read_closer.go
spiffcs: Awesome - was able to reproduce on my end with the new change. TY for the walk through. change is 🟢 for me.

PR #500, opened 2026-01-12
migrate to github.com/moby/moby/client module

thaJeztah:
- [x] stacked on https://github.com/anchore/stereoscope/pull/499


thaJeztah:
I should note that this drops support for EOL versions of Docker (Docker < v25.0), and only supports API versions v1.44 and up; https://pkg.go.dev/github.com/moby/moby/client@v0.2.1#MinAPIVersion

At least for automatic version negotiation; it does still allow overriding the API version through the `DOCKER_API_VERSION` env-var (through `WithAPIVersionFromEnv`; https://pkg.go.dev/github.com/moby/moby/client@v0.2.1#WithAPIVersionFromEnv)

PR #527, opened 2026-02-20
feat: implement array-based tree for 68-76% memory reduction

matthyx:
Replace map-based tree structure with compact array implementation to dramatically reduce memory overhead for container image file trees.
Real world test with `/usr/bin/time go run cmd/syft/main.go scan gitlab/gitlab-ee:latest -o syft-json=/dev/null` shows a significant memory reduction:
```
65.28user 11.02system 0:52.32elapsed 145%CPU (0avgtext+0avgdata 1764408maxresident)k
672inputs+32outputs (0major+675384minor)pagefaults 0swaps
```
to:
```
82.73user 11.24system 1:10.44elapsed 133%CPU (0avgtext+0avgdata 1478712maxresident)k
0inputs+0outputs (0major+601836minor)pagefaults 0swaps
```

### DIFF
--- a/pkg/filetree/depth_first_path_walker.go
+++ b/pkg/filetree/depth_first_path_walker.go
@@ -89,26 +89,27 @@ func (w *DepthFirstPathWalker) Walk(from file.Path) (file.Path, *filenode.FileNo
 
 		// prevent infinite loop
 		if strings.Count(string(currentPath.Normalize()), file.DirSeparator) >= maxDirDepth {
-			return currentPath, currentNode.FileNode, ErrMaxTraversalDepth
+			return currentPath, currentNode.AsFileNode(), ErrMaxTraversalDepth
 		}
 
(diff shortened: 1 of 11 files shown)


kzantow:
Am I reading the results right that runtime is increased noticeably: `65.28user 11.02system 0:52.32elapsed` -> `82.73user 11.24system 1:10.44elapsed`? This looks like almost a 20 second runtime increase on a 50 second scan, that's fairly substantial. Balancing memory usage is important but runtime is also important, I would prefer not to sacrifice one for the other.

I have had some ideas to improve the filetree and overall processing / memory usage in a number of ways, but haven't been able to focus on this yet. I do appreciate the focus here, but just want to emphasize that stereoscope is fundamental to all our tools, as you probably understand, so we just need to be very careful with any changes here. Have you run the full Syft test suite with this change?

PR #539, opened 2026-03-10
go: upgrade dependencies

xnox:
Upgrade all dependencies, as most users of this module already upgrade
these, hence it is best to test with newer versions too.


wagoodman:
We have some automation in place that handles this regularly, I'm going to close this for now. If there is a specific dependency that you want updated that isn't being covered already feel free to open another PR.

PR #545, opened 2026-03-13
fix: enable Docker API version negotiation

jotka:
## Summary

- Add `client.WithAPIVersionNegotiation()` to the Docker client options in `GetClient()`

## Problem

The Docker client in `internal/docker/client.go` uses `client.FromEnv` but does not enable API version negotiation. This means the client defaults to the API version compiled into the `moby/moby` library (currently `v1.53`).

When stereoscope connects to an older Docker daemon — for example one that only supports API `v1.43` — the daemon rejects requests with:

```
Error response from daemon: client version 1.53 is too new. Maximum supported API version is 1.43
```

This breaks tools like grype when scanning images on hosts running older Docker versions (e.g. Docker 25.x, 26.x, Synology DSM).

## Fix

Adding `client.WithAPIVersionNegotiation()` makes the client ping the daemon on first request and automatically downgrade to the daemon's supported API version. This matches the behavior of the Docker CLI itself and grype's own Docker client in `cmd/grype/cli/commands/completion.go`, which already uses `WithAPIVersionNegotiation()`.

## Test plan

- [x] `go build ./internal/docker/` compiles successfully
- [x] `go test ./internal/docker/` passes
- [ ] Verified scanning works against Docker daemon with API v1.43

### DIFF
--- a/internal/docker/client.go
+++ b/internal/docker/client.go
@@ -17,6 +17,7 @@ import (
 func GetClient() (*client.Client, error) {
 	var clientOpts = []client.Opt{
 		client.FromEnv,
+		client.WithAPIVersionNegotiation(),
 	}
 
 	host := os.Getenv("DOCKER_HOST")


wagoodman:
I'm confused how this could make a difference, this [option is deprecated upstream](https://github.com/moby/moby/blob/client/v0.2.2/client/client_options.go#L335-L346) and [auto-negotiation is on by default](https://github.com/moby/moby/blob/client/v0.2.2/client/client.go#L176-L180). There are conditions that [disable negotiation](https://github.com/moby/moby/blob/client/v0.2.2/client/client.go#L217-L221) though, such as setting `DOCKER_API_VERSION` in the environment (since FromEnv is in the options list, and WithAPIVersionFromEnv disables negotiation when set) -- Here's the [PR](https://github.com/moby/moby/pull/51501) that changed this behavior.

Can you double check your findings? Since the option that you added is a `nop` I feel that there must be another reason for the behavior change when pulling this in (maybe due to a version bump of the docker client itself?).

thaJeztah:
Yes, this option is no longer needed, and now enabled by default; the old (`github.com/docker/docker/client`) module did have it as opt-in, but current versions of `github.com/moby/moby/client` enable it by default, so it's now a no-opt; https://github.com/moby/moby/blob/client/v0.4.1/client/client_options.go#L381-L395

- https://github.com/moby/moby/pull/51501

PR #548, opened 2026-03-19
feat: add multi-platform OCI image support

### LINKED ISSUE #175: Support for image indexes with multiple manifests

jasonpaulos:
Add support to OCI directory image provider for multi-platform OCI images.

Currently when a multi-platform OCI image is loaded, stereoscope errors because it only expects a single-platform image.

With this PR, similar to how multi-platform registry images are handled, the user-provided platform is used to select the correct image, unless there is only 1 image available, in which case it will be selected regardless of platform.

I view this as an enhancement, not a breaking change, since this doesn't change any existing behavior outside of cases which would previously fail.

Unit and integration tests added to verify functionality.

This fixes #175 and https://github.com/anchore/syft/issues/1545 and addresses part of https://github.com/anchore/syft/issues/1683. To clarify, this doesn't add support for scanning multiple platforms simultaneously, but it does allow stereoscope & syft to operate on a single image within a multi-platform OCI image, which is currently not possible.


pkg/image/oci/directory_provider.go
```diff
@@ -16,16 +17,24 @@ const Directory image.Source = image.OciDirectorySource
 
 // NewDirectoryProvider creates a new provider instance for the specific image already at the given path.
 func NewDirectoryProvider(tmpDirGen *file.TempDirGenerator, path string) image.Provider {
+	return NewDirectoryProviderWithPlatform(tmpDirGen, path, nil)
+}
+
+// NewDirectoryProviderWithPlatform creates a new provider instance for the specific image already at the given path,
+// with the given platform information to use when loading a multiplatform image.
+func NewDirectoryProviderWithPlatform(tmpDirGen *file.TempDirGenerator, path string, platform *image.Platform) image.Provider {
```
jasonpaulos: I added a new constructor to avoid a potential breaking change on the existing one. Let me know if that's not a concern

jasonpaulos:
@kzantow @wagoodman just checking in on this PR. Let me know if this is something you're open to accepting of if there's additional changes you need to see

jasonpaulos:
@nurmi @spiffcs @kzantow just following up on this PR. This is a blocker for our multi-platform image story at the moment. Please let me know if you are open to accepting this or want to see additional changes. Thanks

kzantow:
Overall this looks great _to me_ -- thanks for the contribution! It looks like there are integration test failure(s). Would you be able to fix that up and potentially open a Syft PR with this module replaced to your branch to see the test results there? That would be helpful to me, at least.

jasonpaulos:
Hi @kzantow, thanks for considering this.

The tests were flaky based on the version of skopeo used to convert the busybox image to an OCI layout. I've fixed that by using a source image that is already in an OCI layout.

I also opened https://github.com/anchore/syft/pull/5074 which references my branch to confirm syft is working properly with this change.

wagoodman (approved):

PR #612, opened 2026-06-04
fix: close layer readers to not block on registry limiter

spiffcs:
## Fix: layer reader leaks which causes a hang in image reads

On `main` `Layer.uncompressedCache` never closed the reader returned by `layer.Uncompressed()`

With go-containerregistry v0.21.6 this leak turns into a hang with image reads containing more than 4 layers.

`v0.21.6` of `go-containerregistry` introduced a pullLimiter in `pkg/v1/remote`. This bounds concurrent blob fetches to defaultJobs = 4. 

This PR closes that reader along with an unclosed file handle in `uncompressedCache.` Closing allows the token to be returned to the new limiter fixing the hang.

### Testing
I built grype against this branch and scanned a reported image from https://github.com/anchore/grype/issues/3492
`go run cmd/grype/main.go openjdk:27-ea-trixie -vvv --from registry` now passes with this patch on my local.

### TBD
There might be some further discussion on configuring/exposing the default concurrency from `go-containerregistry`

Currently it's set at 4: 
https://github.com/google/go-containerregistry/blob/6849394e8a65e2da1177d5b5b00e903551ada1c9/pkg/v1/remote/options.go#L105-L111

### DIFF
--- a/pkg/image/layer.go
+++ b/pkg/image/layer.go
@@ -116,11 +116,13 @@ func (l *Layer) uncompressedCache(uncompressedLayersCacheDir string) (string, er
 	if err != nil {
 		return "", err
 	}
+	defer rawReader.Close()
 
 	fh, err := os.Create(path)
 	if err != nil {
 		return "", fmt.Errorf("unable to create layer cache dir=%q : %w", path, err)
 	}
+	defer fh.Close()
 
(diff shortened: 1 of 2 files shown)


kzantow (approved):
Quick work! 🎉

PR #613, opened 2026-06-05
Add close check ssa linter

wagoodman:
Fixes a few `io.Closer` leaks where we opened something and never closed it. This is the same shape as the go-containerregistry pull-limiter bug ([#2308](https://github.com/google/go-containerregistry/pull/2308)) where a leaked reader holds a slot and deadlocks pulls on multi-layer images. 

Golangci-lint didn't catch these because `bodyclose` only knows about `http.Response.Body`, and nothing in our set flags a generic dropped closer.

The actual leaks fixed:

- `layer.go` `uncompressedCache`: leaked the `Uncompressed()` reader and the cache file handle (#612  fixes this, but I reverted shortly to confirm it would catch this)
- `daemon_provider.go`: never closed the `ImagePull` response body
- `oci/tarball_provider.go`: never closed the opened tarball
- `internal/podman/ssh.go`: never closed the known_hosts file

To stop this from regressing again, adds a small SSA-based analyzer (`test/linter/closecheck`) that follows a closer through the def-use graph and flags it if it's neither closed nor allowed to escape (returned/stored/captured). 

Unlike a syntactic rule (AST) it doesn't false-positive on closers that are returned a few lines later or stashed in a struct, and it caught all of the above. It skips `_test.go` by default and the stdlib-owned `exec.Cmd` pipes, honors `//nolint:closecheck`, and runs as part of `make static-analysis` and `make lint-fix`.


kzantow (approved):
Based on the findings, this looks like what I was expecting 👍

Issue #197, opened 2023-09-08
Support for scanning images built using buildah


hkadakia:
**What would you like to be added**:
Currently syft & grype look for certain locations for finding the container images. It works fine if the container image is built either via `docker` or `podman` as the images they build reside in certain locations and those are part of the supported sources [here](https://github.com/anchore/syft#supported-sources).

Images built using buildah though reside in different location (usually `containers-storage`) and when scanning using syft, it scans the remote image instead of the locally built image. 

**Why is this needed**:
Need to add support for scanning images built using non-daemonize tool like `buildah`. Add a new source where syft can look for when scanning image locally before checking OciRegistry.

**Additional context**:

willmurphyscode:
I just wanted to say this is still on our backlog, and we'd welcome contributions for it, but we haven't been able to prioritize it yet.

bitfisher:
+1

vit-zikmund:
This would be awesome, using the sockets / `image save` is also very inefficient regarding data moving. With containers-storage these are readily accessible using the [containers/storage Go library](https://github.com/containers/storage/tree/main), sometimes exposed into CLI with commands like `podman image mount`, which don't copy any  data at all ❤

Bru456:
Created a PR for this 🙂 #635 any thoughts and feedback back would be great!

Bru456:
PR merged in. @wagoodman This issue can be closed. Thanks for help in the PR!

PR #635, opened 2026-06-29
Adding container-storage as a provider for daemonless Podman and Buildah Support

### LINKED ISSUE #1173: Analyze with Non-Daemon podman
I could be nice if podman:yourrepo/yourimage:tag work with a non-daemonic usage of podman.

**Why is this needed**:
In CICD with rootless setup of buildah/podman there no daemon running.
To make an analysis we have to push first to the docker registry and use registry:yourrepo/yourimage:tag
I could be a waste of time and resources (especially in case how we make an syft + grype and it's useless to push a vulnerable image into the registry).

**Additional context**:

I don't know if it could help but it's ok with dive : https://github.com/wagoodman/dive

Bru456:
Adding Podman daemonless and Buildah support to Stereoscope for issue #197 and Syft issue 2107: [Add support for containers-storage backend for container images](https://github.com/anchore/syft/issues/2107)

### This feature adds: 
* New provider `containers-storage`: 
  * Takes the current user scope (root or user) to determine the container store and copies to a temporary docker archive via `go.podman.io/image/v5` and delegates to the existing Docker archive provider.
  * Ensures the `tags`, `os`, `architecture`, and `variant` is populated.
  * Graceful continuation on miss/unavailable store to allow auto detection to move to the next provider.
* New source `containers-storage` to allow for either auto detection or explicit usage via `--from containers-storage`
* Provider registration: inserted before OCI registry provider so flow is `daemon` -> `containers-storage` -> `remote registry`. Allowing locally built images to resolve before a remote pull.
  * This only applies to Stereoscope I believe Syft and other consumers may have their own ordering.

### Build

The new provider is opt-in at build time. Use the build tag `containers_image_openpgp`.

Without this tag, Stereoscope compiles a stub provider that reports that containers-storage support was not included.

Static Linux builds may also require the `exclude_graphdriver_btrfs` and `exclude_graphdriver_devicemapper` tags. These tags exclude optional graph drivers that require native system dependencies.

### Dependency management
Pulled in `go.podman.io/image/v5 v5.40.0` + `go.podman.io/storage v1.63.0`. 

### Tests
New unit tests are in and passing. 
Tested compiling for both Linux and non Linux builds and bundling this with a local copy of the latest Syft. Docker, Podman (daemon and daemonless) and buildah all working and creating consistent SBOMs. Compared with Syft v1.46.0 
[previous-syft-json-sboms.zip](https://github.com/user-attachments/files/29467314/previous-syft-json-sboms.zip)
 and new version with stereoscope 
[new-syft-json-sboms.zip](https://github.com/user-attachments/files/29467327/new-syft-json-sboms.zip)

### Issues encountered and fixes
* **Dependency compatibility:** The original implementation used the retired `github.com/containers/image/v5` and `github.com/containers/storage` modules and required `go.mod` replacements for `filepath-securejoin` and `selinux`.
* Empty source.metadata (tags / os / architecture): the docker-archive copy is tagless, and the archive provider never sets os/arch. 
  * Fix: gather tags from the store image's names and os/arch/variant from image inspect, and pass them through as additional metadata.
  
  **This has been edited since orginal PR**


go.mod
wagoodman, line 1: This is pulling in github.com/containers/storage , however, this has [recently migrated](https://blog.podman.io/2025/08/upcoming-migration-of-three-containers-repositories-to-monorepo/) to [podman-container-tools/container-libs](https://github.com/podman-container-tools/container-libs).

Additionally, the use of replace directives will most likely block this from being merged as it will complicate / add friction to consumers of stereoscope (e.g. syft and grype and all lib consumers of those projects too).

Bru456: Thanks, @wagoodman I will look into this. The replace was only needed for Linux builds and was needed to get the containers storage to compile. I will see if the new lib sorts the need for replaces. :)

Bru456:
@wagoodman, I’ve migrated the provider to `go.podman.io/image/v5 v5.40.0` and `go.podman.io/storage v1.63.0`, removed both `replace` directives, and updated the imports and documentation accordingly. I have also rebased the branch from the latest commit (at time of writing) from main.

I also validated the changes using a personal Azure DevOps E2E pipeline. The pipeline:

* Cloned and built my stereoscope branch.
* Ran the unit tests with both the default and containers-storage build configurations.
* Cloned the current Syft source and built it against my stereoscope branch without any downstream dependency replacements.
* Built local images using Buildah.
* Successfully scanned those images using both implicit daemonless resolution and explicit `--from containers-storage` resolution.

Both E2E resolution paths completed successfully. Thanks for pointing me toward the migrated modules. 😄

Please note that `containers-storage` support is currently opt-in for downstream consumers such as Syft. Binaries must be built with the `containers_image_openpgp tag`; Linux static builds may also require `exclude_graphdriver_btrfs` and `exclude_graphdriver_devicemapper`.

If keeping this opt-in behavior is preferred, I can follow this with a Syft PR that enables the required tags in its release builds. If you would prefer containers-storage support to be compiled in by default, please let me know and I can revise the approach here.

kzantow:
Hi @Bru456 -- we talked about this briefly on our weekly livestream and I think we're all happy to see this change and it may be pretty close. It looks like some commits need Sign-off and there's a simple conflict to resolve, would you be able to update this PR to address the DCO issue so we can get tests run and give it a thorough review?

Bru456:
@kzantow sorry missed one commit earlier from the sign off :) 

I have rebased and pushed. Let me know if you need anything!

wagoodman:
Wanted to summarize my findings about the Podman dependencies we're pulling in, specifically the CGO story (used `go list -deps -tags containers_image_openpgp`, looking at `.CgoFiles`):
- `github.com/mattn/go-sqlite3` is used for blob-info caching (`go.podman.io/image/v5/copy`), which requires CGO, however, this degrades gracefully to use an in-memory cache (nothing breaking here).
- `github.com/miekg/pkcs11` (from via `containers/ocicrypt`) only exercised for image encryption/decryption which we don't use, so this is not a functional dependency.
- `go.podman.io/storage` has a btrfs graphdriver that is linux only... there is a `dummy_unsupported.go` fallback for `!cgo`.

In terms of building, here's what I saw when using the `stereoscope.ImageProviders` in a test binary:
- `baseline, no tags, CGO_ENABLED=0`: ~ 30 MB
- `-tags containers_image_openpgp, CGO_ENABLED=1`: ~42 MB
- `-tags containers_image_openpgp, CGO_ENABLED=0`: ~39 MB (sqlite and pkcs11 become inert stubs)

To be clear, there's no requirement to use the `containers_image_openpgp` tag, but if someone enables this tag without CGO, there's going to be a silent performance and feature downgrade. 

So the nice thing about this change is that we can now fully support containers-storage backends for overlay (the default driver), vfs, and zfs. It's only when someone is using the btrfs driver where there are opt-in complications.

(I realize that some of this overlaps with what you described in the PR description)

wagoodman:
I'm looking into the CI failures, which I think is partially related to a change that was brought in from the rebase. I'll get another PR going to troubleshoot and fix that, then we'll rebase again.

wagoodman (approved):

PR #647, opened 2026-07-13
file: do not treat a bare .wh. as a whiteout marker

arpitjain099:
`IsWhiteout` returns true for a basename of exactly `.wh.` (the whiteout prefix with nothing after it). `UnWhiteoutPath` then joins the parent directory with an empty name and returns the parent itself, so during squash a crafted layer entry has two effects, both reachable from untrusted image-layer content:

- `/etc/.wh.` deletes the entire `/etc` directory (and all its children) from the squashed tree, instead of a single named sibling.
- a `.wh.` at the image root resolves to `/`, so the merge calls remove on the root path and returns an error that aborts the whole `Squash`. Any consumer (a syft SBOM, a grype scan) then fails to process the image.

The fix requires a non-empty name after the prefix, so a bare `.wh.` is treated as an ordinary file. The opaque marker `.wh..wh..opq` is unaffected. I added a path test for the bare/normal/opaque cases and a squash test showing `/etc/.wh.` no longer wipes `/etc`.

### DIFF
--- a/pkg/file/path.go
+++ b/pkg/file/path.go
@@ -49,7 +49,8 @@ func (p Path) IsDirWhiteout() bool {
 
 // IsWhiteout indicates if the file basename has a whiteout prefix (which means that the file should be removed during squashing)
 func (p Path) IsWhiteout() bool {
-	return strings.HasPrefix(p.Basename(), WhiteoutPrefix)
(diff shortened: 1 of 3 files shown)


wagoodman (approved):
nice find!!

PR #657, opened 2026-07-30
fix: populate OCI platform metadata & do not ignore platform option for single-platform images

jasonpaulos:
This PR makes two improvements when reading OCI directory and tarball images:
1. Populates the image metadata struct with the appropriate platform information. The only case when this is omitted is when no specific platform was requested and the OCI image contains only a single image.
2. When an OCI image contains only a single image, the prior behavior would be to select this image regardless of the requested platform. Now, an error will be thrown if a platform was requested that does not match the single image. If the user does not request a specific platform, the single image will continue to be used regardless of its platform.

I view number 2 as a bugfix, since silently returning an image for a different platform than the one requested is not expected behavior. Additionally, a proper error is already returned if the desired platform isn't available on an OCI bundle that has more than 1 image in it.

### DIFF
--- a/pkg/image/oci/directory_provider.go
+++ b/pkg/image/oci/directory_provider.go
@@ -68,21 +68,25 @@ func (p *directoryImageProvider) Provide(_ context.Context) (*image.Image, error
 	}
 
 	var selectedImage v1.Image
-	if len(allImages) == 1 {
-		// if there is only one image, use it regardless of platform
+	var platformForMetadata *image.Platform
+	if p.platform == nil && len(allImages) == 1 {
+		// if no platform was requested and there is only one image, use it regardless of platform
 		for _, image := range allImages {
 			selectedImage = image.image
 		}
+		platformForMetadata = nil
 	} else {
-		platform := toContainerRegistryPlatform(defaultPlatformIfNil(p.platform))
-		if platform == nil {
+		platform := defaultPlatformIfNil(p.platform)
+		registryPlatform := toContainerRegistryPlatform(platform)
+		if registryPlatform == nil {
 			return nil, fmt.Errorf("error converting platform: %v", p.platform)
 		}
-		matchedImages := imagesForPlatform(allImages, *platform)
+		matchedImages := imagesForPlatform(allImages, *registryPlatform)
 		if len(matchedImages) != 1 {
-			return nil, fmt.Errorf("unexpected number of images matching platform %q in OCI directory (expected 1, found %d)", platform.String(), len(matchedImages))
+			return nil, fmt.Errorf("unexpected number of images matching platform %q in OCI directory (expected 1, found %d)", registryPlatform, len(matchedImages))
 		}
 		selectedImage = matchedImages[0]
+		platformForMetadata = platform
 	}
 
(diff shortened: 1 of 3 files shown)


jasonpaulos:
@wagoodman I just added `--retry-times=3` to the skopeo copy command, which should help with the integration test that just flaked.

wagoodman:
Overall this is the right direction is right. I made a few changes around that observation that platform matching is reading from a different source of truth than what I expected. It decides from the index descriptor, and `platform` there is optional per the spec. `registry_provider.go` already checks the image config instead, and I'd rather follow that one.

I noticed a few things around this:

- descriptors frequently have no `platform` at all, and `findAllImages` only records one when it's present, so the image matches nothing and you get `expected 1, found 0` even when the platform you asked for is correct. Ran `skopeo copy oci:<dir> oci:<dir>` to confirm: the index it writes has no platform field, and pointing the provider at that output with `--platform linux/amd64` fails against an image whose config says linux/amd64. `valid_oci_dir` doesn't catch it because its second entry for the same digest happens to carry a platform. Added `single_image_no_platform_oci_dir` test case to illustrate this.
- `imagesForPlatform` appends the image once per matching platform entry with no break, and `findAllImages` deliberately collects multiple platforms per digest (the comment above it says index entries may reference the same image with different platform info).
- the error is a plain `fmt.Errorf`, but `ErrPlatformMismatch` in `provider.go` exists for exactly this case and the docker daemon, containerd and oci registry providers all return it. Under auto-detect `client.go` folds provider errors into "unable to detect input for X", so a platform mismatch reads like an unreadable file and `errors.As` downstream never fires.
- metadata is built from the platform you passed in rather than the image (but nothing cross-checks the two). The fixture I added for this, `mismatched_descriptor_oci_dir`, has a config saying linux/arm64/v8 behind a descriptor claiming linux/amd64: on current main it happily returns the image and reports amd64. Variant has the same shape in the ordinary case, ask for linux/arm64 against an arm64/v8 image and you record an empty variant.
- for a plain single-arch `oci-dir:` with no platform passed, OS and arch stay empty, which is the "populate OCI platform metadata" half of the title. The test locks it in with `assert.Empty` on `Metadata.Architecture`.

What I changed:

- `readImageMetadata` fills OS/arch/variant from the config it already parses. One spot, every provider gets it, and it reports what the image actually is. Overrides still apply after, so nothing else needed touching, and docker-archive picks it up for free (that provider sets neither today).
- the directory provider takes a single image straight through and then validates against the config on both paths, using `validatePlatform` already in the package. That covers the optional-descriptor case, gets the typed error back, and `platformForMetadata` goes away.
- smaller ones alongside: the `break` in `imagesForPlatform`, skip buildx attestation manifests so they stop counting as images, fix the constructor godoc on both `...WithPlatform` functions since they still say "when loading a multiplatform image", and have skopeo stage to a tmp path so an interrupted copy can't poison the fixture cache.
- fixtures for each of the above. Also swapped the `assert.Error` in both tables for `require.Error`, since the line after it calls `err.Error()` and segfaults the moment a case stops erroring. Hit that while checking these and it silently ate five subtests.

(force pushed to rebase was all)

wagoodman (approved):

PR #665, opened 2026-08-12
fix(filtree): hardlink false cycle

spiffcs:
### Resolve hardlinks to the file they were bound to

A hardlink is a second name for a file that already exists. It is not a path to be looked up later. `FileNode` stored `LinkPath` for hardlinks and re-resolved it during link resolution when considering an images squashed tree. This made the hardlink behave like a symlink.

Package managers like `uv` and `pip` install a file and hardlink it into a content-addressed cache. A later image layer can replace the installed name with a symlink aimed at the cache name. Currently, when considering and images squashed representation, the two names would point at each other and resolution reports "cycle during symlink resolution"

This fix captures the target reference in `AddHardLink`, when the link is created and returns it during resolution instead of following LinkPath.

When the reference is absent the binding is left empty and resolution falls back to the previous behavior.

### DIFF
--- a/pkg/filetree/filenode/filenode.go
+++ b/pkg/filetree/filenode/filenode.go
@@ -9,10 +9,19 @@ import (
 )
 
 type FileNode struct {
-	RealPath       file.Path // all constituent paths cannot have links (the base may be a link however)
-	FileType       file.Type
-	LinkPath       file.Path // a relative or absolute path to another file
-	Reference      *file.Reference
+	RealPath  file.Path // all constituent paths cannot have links (the base may be a link however)
+	FileType  file.Type
+	LinkPath  file.Path // a relative or absolute path to another file
+	Reference *file.Reference
+
+	// HardLinkTarget is the file this node was bound to when the hardlink was created.
+	// a hardlink's link path is always a real path to a file that exists in the same layer,
+	// A symlink's link path is just text; which file it names depends on the layer topology
+	// it's important to note a later layer can replace the destination
+	//
+	// Holding the reference (pointed to the contents at the layer is was created)
+	// rather than only LinkPath (a path with no layer information) is what
+	// lets later callers reach the file the link was bound to, even once another layer has replaced that path.
 	HardLinkTarget *file.Reference
 }
 
--- a/pkg/filetree/filenode/filenode.go
+++ b/pkg/filetree/filenode/filenode.go
@@ -9,10 +9,11 @@ import (
 )
 
 type FileNode struct {
-	RealPath  file.Path // all constituent paths cannot have links (the base may be a link however)
-	FileType  file.Type
-	LinkPath  file.Path // a relative or absolute path to another file
-	Reference *file.Reference
+	RealPath       file.Path // all constituent paths cannot have links (the base may be a link however)
+	FileType       file.Type
+	LinkPath       file.Path // a relative or absolute path to another file
+	Reference      *file.Reference
+	HardLinkTarget *file.Reference
 }
 
 func NewDir(p file.Path, ref *file.Reference) *FileNode {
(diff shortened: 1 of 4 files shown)


pkg/filetree/union_filetree_test.go
wagoodman, line 1: We should probably have more tests here, at least one on the content fetching side (does the content api correctly fetch the contents for the original file).

pkg/filetree/union_filetree_test.go
```diff
@@ -171,3 +174,65 @@ func TestUnionFileTree_Squash_whiteout(t *testing.T) {
...
 }
+
+// uv and pip install can write a  file and then hardlink it into a content-addressed cache,
+// so two names share one inode. A later layer can replace the installed name with a
+// symlink aimed at the cache name:
+//
+//	layer 40:  /opt/venv/.../cli-32.exe    -+-> inode X   (two names, one file)
+//	           /root/.cache/.../cli-32.exe -+
+//	layer 43:  /opt/venv/.../cli-32.exe    ---> symlink --> /root/.cache/.../cli-32.exe
+//
```
wagoodman, line 185: this is a great example, but we should also illustrate this from the tar header perspective, since each layer isn't measured inodes yet until the final filesystem is assembled for the container. 

For instance, we're leaning on the assumption that the RealFile shows up first as a tar header with contents, and that the second tar header is the HardLink, with no contents... this implies other tar header configurations we're assuming we WONT see in the real world: instead of RealFile+contents then HardLink+empty entries, seeing HardLink+contents then HardLink+empty (or flipped), or HardLink+empty then RealFile+contents (flipped from expected), or HardLink+contents then RealFile+empty (just plain weird). It would be nice to bake some of these assumptions into the test file here or maybe in other places in processing.

pkg/filetree/filetree.go
```diff
@@ -372,6 +372,20 @@ func (t *FileTree) resolveNodeLinks(n *nodeAccess, followDeadBasenameLinks bool,
 			return nil, ErrLinkCycleDetected
 		}
 
+		// a hardlink names a file, not a path to be looked up again later. Return the file it was bound to when the
+		// link was created. If we follow LinkPath here could instead re-resolve to a symlink (created at another layer).
+		// In the squashed case this symlink could point back at us causing an incorrect cycle detection
+		if currentNodeAccess.FileNode.FileType == file.TypeHardLink && currentNodeAccess.FileNode.HardLinkTarget != nil {
+			bound := *currentNodeAccess.FileNode
+			bound.Reference = bound.HardLinkTarget
+			currentNodeAccess = &nodeAccess{
```
wagoodman, line 387: I think this is right, but this is worth thinking about if the top level API might raise up something that looks strange. For instance we'd have a HardLink type with a size of 0 but with non-zero contents. I think there is good argument to be made that this is really a figment of the strangeness of tar behavior and in reality stereoscope is just raising up what actually exists.

pkg/filetree/filenode/filenode.go
```diff
@@ -9,10 +9,11 @@ import (
...
 type FileNode struct {
-	RealPath  file.Path // all constituent paths cannot have links (the base may be a link however)
-	FileType  file.Type
-	LinkPath  file.Path // a relative or absolute path to another file
-	Reference *file.Reference
+	RealPath       file.Path // all constituent paths cannot have links (the base may be a link however)
+	FileType       file.Type
+	LinkPath       file.Path // a relative or absolute path to another file
+	Reference      *file.Reference
+	HardLinkTarget *file.Reference
```
wagoodman, line 25: I think the reason why we can do this for hardlinks and not symlinks is because the link paths for hardlinks are always realpaths for things that always exist within the same layer. This is not true of symlinks, which the layer topology changes the answer about what the link might point to. This is probably worth capturing as a comment here as it seems odd, even redundant, to capture the HardLinkTarget in this way... but there is a good but subtle reason too: it locks in the specific file reference (path at a layer) instead of just the LinkPath which lacks layer information for future callers.

pkg/filetree/union_filetree_test.go
spiffcs: I added more tests that fail if we revert to the content APIs in pkg/image/image_test.go

spiffcs:
Closing this in favor of #670 after doing review with @wagoodman and talking about the different approaches.

PR #666, opened 2026-08-13
fix: continue file tree searches after link cycles

### LINKED ISSUE #206: Cycle during symlink resolution prevents syft/grype scans from enumerating vulnerabilities
**What happened**:
We recently came across an image where a user had accidentally created a symlink loop within their container image.  This resulted in the image failing both syft and grype scans.

This could be used by an attacker or developer who wants to hide vulnerabilities through [malicious compliance](https://github.com/bgeesaman/malicious-compliance).  By generating a symlink loop, syft/grype will error and fail to output results.  If scan errors are not closely monitored the image could avoid detection.

Malformed symlinks should be logged, but allow the rest of the syft or grype scans to complete.

**How to reproduce it (as minimally and precisely as possible)**:
Using old, known vulnerable image from `webgoat/webgoat-8.0:latest` grype shows a bunch of vulns:
```
% grype webgoat/webgoat-8.0:latest
 ✔ Vulnerability DB                [no update available]
 ✔ Pulled image
 ✔ Loaded image                                                                                                                                                             webgoat/webgoat-8.0:latest
 ✔ Parsed image                                                                                                                sha256:6664051b8808540cf920e5802e7eb025f9cb19346dcb0fc2be137a26979bb111
 ✔ Cataloged packages              [315 packages]
 ✔ Scanned for vulnerabilities     [581 vulnerability matches]
   ├── by severity: 73 critical, 185 high, 152 medium, 60 low, 111 negligible
   └── by status:   304 fixed, 277 not-fixed, 0 ignored
```

Build a downstream image and create a symlink loop in a cataloged binary, `/usr/bin/xz` for example:
```
echo "FROM webgoat/webgoat-8.0:latest
USER root
RUN yes | ln -sfi /usr/bin/xzcat /usr/bin/xz
USER webgoat" | docker build -t symlink-loop:latest .
```

Syft error:
```
% syft symlink-loop:latest
 ✔ Loaded image                                                                                                                                                                    symlink-loop:latest
 ✔ Parsed image                                                                                                                sha256:108309a3bb9c201ffe0f2c4fc1300f434485e5216d6edad8c76d6d2ec8d3e7de
 ✔ Cataloged packages              [315 packages]
[0003]  WARN unable to create any package-file relationships cataloger=dpkgdb-cataloger error=unable to find path for path="/usr/bin/xz": cycle during symlink resolution package=xz-utils
[0003]  WARN unable to process mimetypes=[application/x-executable application/x-mach-binary application/x-elf application/x-sharedlib application/vnd.microsoft.portable-executable]: unable to get ref
[0005]  WARN error while cataloging cataloger=graalvm-native-image-cataloger
1 error occurred:
        * failed to find binaries by mime types: unable to get ref for path="/usr/bin/xz": cycle during symlink resolution
```

Grype error:
```
% grype symlink-loop:latest
 ✔ Vulnerability DB                [no update available]
 ✔ Loaded image                                                                                                                                                                    symlink-loop:latest
 ✔ Parsed image                                                                                                                sha256:108309a3bb9c201ffe0f2c4fc1300f434485e5216d6edad8c76d6d2ec8d3e7de
 ✔ Cataloged packages              [315 packages]
[0003]  WARN unable to create any package-file relationships cataloger=dpkgdb-cataloger error=unable to find path for path="/usr/bin/xz": cycle during symlink resolution package=xz-utils
[0003]  WARN unable to process mimetypes=[application/vnd.microsoft.portable-executable application/x-executable application/x-mach-binary application/x-elf application/x-sharedlib]: unable to get ref
[0004]  WARN error while cataloging cataloger=graalvm-native-image-cataloger
1 error occurred:
        * failed to catalog: 1 error occurred:
        * failed to find binaries by mime types: unable to get ref for path="/usr/bin/xz": cycle during symlink resolution
```

This issue was validated on latest syft/grype versions:
```
% syft version
Application: syft
Version:    0.94.0
BuildDate:  2023-10-20T17:21:07Z
GitCommit:  8f6bdde6662aa8050a71eadbdb7bd5a3b079a56d
GitDescription: v0.94.0
Platform:   linux/amd64
GoVersion:  go1.21.3
Compiler:   gc

% grype version
Application:         grype
Version:             0.72.0
BuildDate:           2023-10-20T18:17:05Z
GitCommit:           04df28051b7694a5e4a28fc5b2ea2068f24ef213
GitDescription:      v0.72.0
Platform:            linux/amd64
GoVersion:           go1.21.3
Compiler:            gc
Syft Version:        v0.94.0
Supported DB Schema: 5
```

**Environment**:
- OS (e.g: `cat /etc/os-release` or similar):
```
$ cat /etc/os-release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.9 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="7.9"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.9 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.9:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.9
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.9"
```

CAOShurong:
## Summary

- continue depth-first traversal after a path fails specifically with `ErrLinkCycleDetected`
- skip the same malformed entries during indexed file-tree searches instead of aborting all results
- warn with the affected path while preserving failures for every other error type
- add regressions proving files after a two-link cycle remain traversable and MIME-searchable

Fixes #206.

## Validation

- `go test ./pkg/filetree -count=1`
- `go test ./pkg/filetree -run 'TestDFS_WalkAll_SkipsLinkCycles|Test_searchContext_SearchByMIMEType_SkipsLinkCycles' -count=20`
- `go vet ./...`
- standalone API probe using `/usr/bin/xz <-> /usr/bin/xzcat`: traversal and MIME search both return the normal files before and after the cycle with no error

The broader `go test ./...` run was also attempted on Windows. The changed `pkg/filetree` package passed; unrelated platform/integration suites retain existing Windows assumptions and require a Docker daemon, so I have not represented that command as passing.

## AI assistance

OpenAI Codex assisted with issue screening, implementation, and test drafting. I reviewed the resulting diff and ran the validations listed above. The commit includes an `Assisted-by` trailer and DCO sign-off.


wagoodman:
I think the approach taken here was correct, I took it a little further.

The same abort existed in most of the other places that resolve links, so the commits on top of yours extend the skip to `FilesByGlob`, `ListPaths`, `SearchByPath`, the sub-directory search, the backward reference index, `Merge`, and the two resolve helpers in `pkg/image`.

What decides whether a given call site gets the skip: resolving one specific path reports the error, since the caller asked about exactly that path. Enumerating many paths skips the bad one and keeps going, since one broken link should not cost the caller every other result. `FileTree.File` still errors for that reason.

Two of the remaining sites were worse than the original bug:

- `Merge` resolves through links when an upper layer has an opaque directory, so a cycle in a lower layer failed the squash, which fails the whole image load. Not one file, the entire read.
- the backward reference index swallowed its error into a warning already, so it silently emptied itself rather than aborting. Files reachable through symlinked directories just went missing from results. This was a little worse than a loud failure, and it needed guarding in more places was all.

I made more changes as well:

- the skip now covers `ErrLinkResolutionDepth` as well. A chain too long to follow denies a scan the same way a cycle does. This expands the usecase beyond just cycles only.
- `IsUnresolvableLink` is exported. That is the only public API addition here, and it exists because the classification was already being copy-pasted into `pkg/image` and the `Merge` fix would have made a third copy.

I'll rebase and push up shortly...

wagoodman (approved):

PR #670, opened 2026-08-17
fix(image): bind hardlinks to the file they name when indexing a layer

wagoodman:
This is an alternative to #665 that came up during its review.

A hardlink is a second name for a file that already exists, not a path to look up later. We store only the link path and re-resolve it against the squashed tree, which makes a hardlink behave like a symlink and gets a few things wrong:

- two names that end up naming each other read as a link cycle. This is what `uv` and `pip` produce (install a file, hardlink it into a content-addressed cache, then a later layer replaces the installed name with a symlink aimed at the cache name) and it aborts resolution for anything walking files
- if a later layer overwrites or replaces the *other* name, we report that layer's contents for this one
- if a later layer deletes the other name, this one reads as missing, though the container serves it fine
- the entry reports size 0 and no MIME type, since a hardlink tar header has no data section, so hardlinked binaries are silently skipped by anything enumerating by MIME type

The fix binds the name to the file it refers to when the layer tar is read instead of re-resolving during path resolution, so `pkg/filetree` is untouched. The catalog is keyed by `file.ID` and is never squashed or merged, so a later layer can't change what a name refers to, which is the guarantee the filesystem actually gives. 

One visible change worth calling out: a hardlinked name is now catalogued as the file it names, so `file.TypeHardLink` no longer appears for well-formed images and such a name resolves to itself rather than to the other name. On disk neither name is "the hardlink", so this is closer to the truth and it's what [syft #5029](https://github.com/anchore/syft/issues/5029) wanted from the other side, but it does mean nothing can answer "is this path also known as something else" anymore.

Tests come in two layers, a real `docker build` fixture over 11 hardlink shapes across every image source plus a docker-free equivalent in the unit suite, and the first commit is deliberately red so the broken cases are visible before the fix lands.

### DIFF
--- a/pkg/filetree/filetree.go
+++ b/pkg/filetree/filetree.go
@@ -554,7 +554,7 @@ func (t *FileTree) AddHardLink(realPath file.Path, linkPath file.Path) (*file.Re
 	if fna.HasFileNode() {
 		// this path already exists
 		if fna.FileNode.FileType != file.TypeHardLink {
-			return nil, fmt.Errorf("path=%q already exists but is NOT a symlink file", realPath)
+			return nil, fmt.Errorf("path=%q already exists but is NOT a hardlink file", realPath)
 		}
 		// this is a symlink file, provide a new or existing file.Reference
 		if fna.FileNode.Reference == nil {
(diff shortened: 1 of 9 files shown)


spiffcs (approved):
@wagoodman I'm good with this over my original approach. I noticed this is in draft. Is there anything else you wanted me to help add?

wagoodman:
I want to spend more time with this... I think it's the right direction, but not sure it's complete yet.

PR #680, opened 2026-08-31
perf: take layer diff IDs from the image config instead of recomputing

asomya:
## Issue

A layer's DiffID is the digest of its *uncompressed* content. For an OCI layout, ggcr's `layer.DiffID()` gunzips the entire blob just to hash it; stereoscope asked for the DiffID while building layer metadata and then gunzipped the same blob again to unpack it — every byte of every layer decompressed twice. The waste is complete: the image config already lists every layer's DiffID in `rootfs.diff_ids`.

## Fix

`Image.Read` reads `rootfs.diff_ids` from the image config once and passes the known DiffID into each layer's metadata (`newLayerMetadata`). Fallback: when the config is unavailable or its diff-ID count does not line up with the layer list, the DiffID is computed exactly as before, so no input regresses.

## Measured

Imported into anchorectl and compared back to back with the same anchorectl commit on current stereoscope main (the `central-ocidir` scenario is an OCI layout on disk — the case this targets), medians of 3:

|---|---|---|
| enterprise:dev from the docker daemon (1.28 GB, 40k files) | 15.83 s / 22.33 s / 948 MB | 15.41 s / 22.24 s / 941 MB (-3% wall) |
| nginx from a registry (loopback) | 1.55 s / 4.73 s / 246 MB | 1.53 s / 4.68 s / 241 MB (-1% wall) |
| enterprise:dev from a registry (loopback) | 9.41 s / 24.00 s / 973 MB | 9.37 s / 23.96 s / 935 MB (-0% wall) |
| OCI layout on disk (centralized analyze path) | 14.15 s / 29.12 s / 966 MB | 9.01 s / 24.31 s / 1021 MB (-36% wall) |

The OCI-layout scenario is the target and drops 36% wall / 17% CPU (the redundant gunzip of every layer). Other sources already knew their DiffIDs cheaply and are unchanged within noise. An interleaved A/B re-check of the RSS delta on the OCI scenario showed +3.3%, within run-to-run noise.


wagoodman:
A couple things needed fixing first (I already pushed adjustments):

- The diff ID isn't just a label we report, it's also the filename we cache the unpacked layer under. Once it comes from the config, an image claiming the same diff ID for two layers makes the second one read the first's content, and its own files quietly vanish (no error and nothing logged))
- The doc on the exported digest field still said it was the sha256 of the layer contents. It's now what the image claims about itself. Same value for any well formed image, but it's a semantics change on something that lands in SBOMs as layer identity, so it wants a release note

What I pushed:

- Rebased onto main first, since the layer descriptor work that landed rewrote the same loop
- Cache filename picked up the layer index so layers can't collide
- Dropped the redundant config read in favor of the copy we already have
- Digest doc now says what the value actually is
- Tests proving the config path gives the same answer as computing it, and that the fallback works. Kept your panicking DiffID stub, nice way to show we never touch the layer

One for you to check: the "config unavailable, fall back" case can't actually happen, we bail out earlier when the config fails to load... so I dropped the tests covering it, which reverted part of your last commit (but did keep your panicking DiffID stub).

wagoodman (approved):

PR #681, opened 2026-08-31
perf: overlap layer fetch and indexing in Image.Read

asomya:
## Issue

`Image.Read` handled layers strictly one at a time: download a layer, decompress it, walk its tar, then start the next. Downloading is network work, decompressing/indexing is CPU work, and running them back to back keeps exactly one resource busy at any moment — total read time is the **sum** of network and CPU time when it could approach the **max** of the two.

## Fix

- `Layer.Read` = `Layer.Fetch` (download + decompress into the layer cache) followed by `Layer.Index` (tar walk, file-tree build); `Image.Read` drives the two stages as a pipeline with separate worker pools.
- `WithLayerReadConcurrency(n)` bounds concurrent fetches per image; indexing always overlaps fetching.
- **Source-aware concurrency, set from measurement:** the registry provider fetches **one layer at a time** — against nvcr.io on a 10.5 GiB image, one stream ran ~62 MB/s while four in parallel managed ~27 MB/s aggregate (pull 3m20s → 6m18s) and even two streams were ~20% slower, because a single stream already saturated the link. Local sources (daemon tarballs, OCI layouts) are disk-bound and decompress several layers at once (default NumCPU, capped at 8; `LayerReadConcurrency` overrides).
- Cache-path population is serialized per digest, so two layers sharing a digest (empty layers do) cannot race under concurrent fetch.

## Measured

Imported into anchorectl, back to back with the same anchorectl commit on current stereoscope main, medians of 3:

|---|---|---|
| enterprise:dev from the docker daemon (1.28 GB, 40k files) | 15.83 s / 22.33 s / 948 MB | 16.02 s / 22.48 s / 918 MB (+1% wall) |
| nginx from a registry (loopback) | 1.55 s / 4.73 s / 246 MB | 1.45 s / 4.69 s / 245 MB (-6% wall) |
| enterprise:dev from a registry (loopback) | 9.41 s / 24.00 s / 973 MB | 9.31 s / 23.94 s / 922 MB (-1% wall) |
| OCI layout on disk (centralized analyze path) | 14.15 s / 29.12 s / 966 MB | 13.50 s / 29.60 s / 960 MB (-5% wall) |

The daemon scenario is dominated by a single-stream `docker save`, so wall stays roughly flat there; loopback-registry wall is unchanged by design (fetch stays at 1) with indexing overlap giving the small wins shown. The gains grow with layer count and slower disks; the registry single-fetch rationale carries its own measurements (see commit message).

## Ordering guarantee

Layers may fetch and index in any order, but nothing downstream can observe that: results land in `layers[idx]` by manifest position, `readLayers` returns only after **every** layer is indexed, and only then does `Image.Read` run the existing bottom-up squash — so squashed trees are still built lowest layer first, over a fully indexed, correctly ordered layer list.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,7 +37,10 @@ require (
 	golang.org/x/crypto v0.55.0
 )
 
-require golang.org/x/tools v0.49.0
+require (
+	github.com/anchore/go-sync v0.1.2
+	golang.org/x/tools v0.49.0
+)
 
 require (
 	cloud.google.com/go/compute/metadata v0.9.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -18,6 +18,8 @@ github.com/anchore/go-homedir v0.1.1 h1:KpaO5BPrVT7cVMcEZ9KfFfEi0iuo3DtagPGl0C74
 github.com/anchore/go-homedir v0.1.1/go.mod h1:9B0DGhbmMAJVGGEbrlb+PkORM5eDFCEOtZ3xQ22qaLA=
 github.com/anchore/go-logger v0.1.1 h1:HPBNmQVBdYliwaFVUL7oJ3UvSlaoXAMj/ZagAIYaCpA=
 github.com/anchore/go-logger v0.1.1/go.mod h1:ekWuh5BkZVwyXnyEd4fwL29N4pNLssgRsqh2+kFN/b8=
+github.com/anchore/go-sync v0.1.2 h1:RDD5RCanrWqhbtBulqPSZpERadUlgcYjpybYzYd3Uk8=
+github.com/anchore/go-sync v0.1.2/go.mod h1:grjZH059bCHM6f6jh+Y7VYvX/q8YcjkDJqG6nplHGkE=
 github.com/aws/aws-sdk-go-v2 v1.41.2 h1:LuT2rzqNQsauaGkPK/7813XxcZ3o3yePY0Iy891T2ls=
 github.com/aws/aws-sdk-go-v2 v1.41.2/go.mod h1:IvvlAZQXvTXznUPfRVfryiG1fbzE2NGK6m9u39YQ+S4=
 github.com/aws/aws-sdk-go-v2/config v1.32.10 h1:9DMthfO6XWZYLfzZglAgW5Fyou2nRI5CuV44sTedKBI=
--- a/pkg/filetree/index.go
+++ b/pkg/filetree/index.go
@@ -207,7 +207,11 @@ func (c *index) GetByExtension(extensions ...string) ([]IndexEntry, error) {
 func (c *index) GetByBasename(basenames ...string) ([]IndexEntry, error) {
 	c.RLock()
 	defer c.RUnlock()
+	return c.getByBasename(basenames...)
+}
 
+// getByBasename is the lock-free implementation of GetByBasename. The caller must hold at least a read lock.
+func (c *index) getByBasename(basenames ...string) ([]IndexEntry, error) {
 	var entries []IndexEntry
 
 	for _, basename := range basenames {
@@ -249,7 +253,7 @@ func (c *index) GetByBasenameGlob(globs ...string) ([]IndexEntry, error) {
 		var e error
 		c.basenames.Each(func(b string) bool {
 			if patternObj.IsMatch(b) {
-				bns, err := c.GetByBasename(b)
+				bns, err := c.getByBasename(b)
 				if err != nil {
 					e = fmt.Errorf("unable to fetch file references by basename (%q): %w", b, err)
 					return false
--- a/pkg/filetree/index_test.go
+++ b/pkg/filetree/index_test.go
@@ -3,10 +3,13 @@
 package filetree
 
 import (
+	"fmt"
 	"io/fs"
 	"slices"
 	"strings"
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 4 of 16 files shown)


wagoodman:
I made a bunch of changes to this. The reason being, there was [already a prototype](https://github.com/anchore/stereoscope/compare/main...concurrent-layer-reading) that attempted to tackle adding concurrency here. Your approach and the prototype were a little different:
- your approach (this PR, we'll call option "A") is a two stage pipeline (fetch and index pools) bounded per-image
- the prototype (we'll call option "B") was a single processing pool where the work unit was defined as download+squash for a single layer.

Approach "A" makes sense from the perspective that we're splitting pools based on the resource being bounded (one fore network/disk IO and the other for CPU). This means we can express something like "one download at a time, but index eight layers in parallel".

Approach "B" makes sense because we're able to overlap the squash phase with layer reading. This means we're limited to expressing only "N layers end to end at a time". In practice, after characterizing the prototype, it is slower than the serial case in terms of downloading from a registry.

The changes I made I'll call approach "C": Take the pool decomposition from "A" and the squash overlap from "B". Additionally, from an API ergonomics, "B" uses `go-sync` and other primitives that we've settled on for expressing concurrency in a library in other projects, so that was worth carrying forward.

I went to characterize the performance of these flavors, but keep in mind there is yet another aspect that interferes with measuring the results: #679 was merged while working on this... but since it removed several syscalls from the layer processing path all measurements must be relative to after that change, not before it.

I'll post the performance comparison shortly...

wagoodman:
I tried out a few different fixtures:

| id | layers | uncompressed | file entries | shape |
|----|-------:|-------------:|-------------:|-------|
| f1-few-fat | 5 | 249 MiB | 537 (99 regular, 334 symlink) | one big fat layer, few files — fetch/decompress bound |
| f2-many-thin | 31 | 26 MiB | 6,577 (6,087 regular) | many thin layers — layer-count parallelism |
| f3-many-files | 7 | 58 MiB | 51,789 (50,687 regular) | few layers, huge fan-out — index bound |
| f4-link-heavy | 9 | 21.6 MiB | 23,718 (3,337 regular, 9,909 symlink, 10,000 hardlink, 150 whiteout, max chain depth 5, 8,571 cross-layer) | squash + link-resolution bound |
| f5-realistic | 11 | 591 MiB | 15,506 (12,730 regular) | real base image + package installs |

The short answer is that A & C are about the same on the "read" phase for wall-clock time:

| fixture | M | A | B | C |
|---|---|---|---|---|
| f1-few-fat | 390 | 235 (1.66x) | 394 (0.99x) | 236 (1.65x) |
| f2-many-thin | 369 | 167 (2.21x) | 375 (0.98x) | 165 (2.23x) |
| f3-many-files | 872 | 496 (1.76x) | 859 (1.01x) | 499 (1.75x) |
| f4-link-heavy | 515 | 361 (1.43x) | 503 (1.02x) | 359 (1.44x) |
| f5-realistic | 3817 | 1376 (2.77x) | 3798 (1.01x) | 1365 (2.80x) |

M is main branch after merging #679, this is showing the wall-clock speedups in the read phase only... where A ~= C.

But the "squashing" phase is being shared more with C:

| fixture | N | A | C | C vs A |
|---|---|---|---|---|
| f4-link-heavy | 1 | 503.5 | 380.7 | 24% faster |
| f2-many-thin | 1 | 307.7 | 253.8 | 18% faster |
| f2-many-thin | 2 | 203.5 | 158.2 | 22% faster |

So in this case C > A.

In terms of memory performance (peak RSS in MB), it's about the same:

| fixture | M | A | B | C |
|---|---|---|---|---|
| f1-few-fat | 18.3 | 19.3 (+5.3%) | 18.5 (+0.9%) | 19.1 (+4.5%) |
| f2-many-thin | 80.9 | 81.0 (+0.2%) | 79.4 (-1.9%) | 79.0 (-2.3%) |
| f3-many-files | 206.6 | 217.2 (+5.1%) | 210.1 (+1.7%) | 207.9 (+0.6%) |
| f4-link-heavy | 163.9 | 162.5 (-0.8%) | 168.5 (+2.9%) | 159.7 (-2.5%) |
| f5-realistic | 111.0 | 116.0 (+4.5%) | 111.5 (+0.5%) | 113.2 (+2.0%) | 

... and importantly it is not growing with concurrency (still peak RSS in MB):

| fixture | N | M | A | B | C |
|---|---|---|---|---|---|
| f2-many-thin | 1 | 79.8 | 81.9 | 81.9 | 81.9 |
| f2-many-thin | 2 | 80.0 | 78.1 | 81.0 | 79.1 |
| f2-many-thin | 4 | 79.7 | 79.1 | 78.8 | 79.5 |
| f2-many-thin | 8 | 77.5 | 80.5 | 77.4 | 77.4 |
| f2-many-thin | 12 | 78.7 | 82.2 | 81.6 | 80.2 |
| f4-link-heavy | 1 | 167.9 | 163.4 | 164.5 | 164.6 |
| f4-link-heavy | 2 | 162.1 | 160.1 | 157.7 | 158.5 |
| f4-link-heavy | 4 | 166.2 | 159.9 | 163.7 | 159.4 |
| f4-link-heavy | 8 | 165.0 | 160.6 | 166.2 | 161.4 |
| f4-link-heavy | 12 | 160.2 | 167.1 | 165.9 | 164.3 |

So peak memory is being determined by the squashing of the layer trees, not by how many layers are being processed at a time.

kzantow (approved):

PR #682, opened 2026-08-31
fix: make layer-cache writes atomic and serialized per path

asomya:
## Issue

The uncompressed-layer cache wrote each decompressed layer directly to its final digest-keyed path (`os.Create` + `io.Copy`). Two writers of the same digest can interleave, and any reader arriving mid-write sees a truncated layer — silent corruption rather than an error. Two layers of one image *can* share a digest (empty layers do), and any future concurrency around layer reads or a longer-lived cache directory makes this a real race.

## Fix

- Write to `os.CreateTemp` in the cache directory and `os.Rename` into place — atomic on POSIX filesystems, so a cache path either does not exist or holds a complete layer, never a partial one.
- A per-path mutex serializes writers of one digest; the loser of the race finds the winner's completed file on the `os.Stat` check and reuses it instead of rewriting.
- Failures remove the temp file, so nothing half-written is left behind under any name.

**Why in-process mutexes and not lock files:** an on-disk lock has a liveness problem — if the process holding it dies, the next process cannot tell whether the holder is alive. The per-path locks here are plain `sync.Mutex`es that die with the process, so a stale lock cannot exist; cross-process safety comes from the write protocol instead (a writer only exposes a complete layer via rename, and a crash leaves at most an orphaned `*.partial` that no reader looks at).

Unit test covers concurrent writers of one path: no overlap, correct content, no `.partial` files left.

## Measured

Imported into anchorectl and compared back to back with the same anchorectl commit on current stereoscope main (expected to be performance-neutral; the point is correctness), medians of 3:

|---|---|---|
| enterprise:dev from the docker daemon (1.28 GB, 40k files) | 15.83 s / 22.33 s / 948 MB | 15.81 s / 22.12 s / 970 MB (-0% wall) |
| nginx from a registry (loopback) | 1.55 s / 4.73 s / 246 MB | 1.54 s / 4.70 s / 246 MB (-1% wall) |
| enterprise:dev from a registry (loopback) | 9.41 s / 24.00 s / 973 MB | 9.37 s / 23.90 s / 929 MB (-0% wall) |
| OCI layout on disk (centralized analyze path) | 14.15 s / 29.12 s / 966 MB | 14.51 s / 30.15 s / 939 MB (+3% wall) |

Performance-neutral within run-to-run noise, as expected — this PR is a correctness fix; the numbers are here to show it costs nothing.

### DIFF
--- a/pkg/image/layer.go
+++ b/pkg/image/layer.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path"
 	"strings"
+	"sync"
 	"time"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
@@ -118,6 +119,11 @@ func (l *Layer) uncompressedCache(uncompressedLayersCacheDir string) (string, er
 	// layer's unpacked content.
 	path := path.Join(uncompressedLayersCacheDir, fmt.Sprintf("%d-%s", l.Metadata.Index, l.Metadata.Digest))
 
+	// two layers of one image may share a digest (empty layers do); only one populates the cache
+	// path, anyone else waits and then finds it complete
+	unlock := lockCachePath(path)
+	defer unlock()
+
 	if _, err := os.Stat(path); !os.IsNotExist(err) {
 		return path, nil
 	}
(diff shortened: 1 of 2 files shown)


pkg/image/layer.go
```diff
@@ -118,20 +124,45 @@ func (l *Layer) uncompressedCache(uncompressedLayersCacheDir string) (string, er
...
 		return "", fmt.Errorf("unable to populate layer cache dir=%q : %w", path, err)
 	}
+	if err := fh.Close(); err != nil {
+		_ = os.Remove(tmpPath)
+		return "", fmt.Errorf("unable to finish layer cache dir=%q : %w", path, err)
+	}
+	if err := os.Rename(tmpPath, path); err != nil {
+		_ = os.Remove(tmpPath)
+		return "", fmt.Errorf("unable to place layer cache dir=%q : %w", path, err)
+	}
```
wagoodman: there is a more idiomatic way to deal with the cleanup, an adhoc defer:
```
defer func() {
  fh.Close()
  os.Remove(tempPath)
}()
```

pkg/image/layer.go
```diff
@@ -118,20 +124,45 @@ func (l *Layer) uncompressedCache(uncompressedLayersCacheDir string) (string, er
...
 	return path, nil
 }
 
+// cachePathLocks serializes population of one cache path across layers read concurrently. The
+// locks are in-process mutexes on purpose: they die with the process, so a crash can never leave a
+// stale lock a later run would have to reason about (the liveness problem on-disk lock files have).
+// Cross-process safety comes from the write protocol instead - a writer only ever exposes a
+// complete layer via rename, and a crash leaves at most an orphaned *.partial temp file that no
+// reader ever looks at.
+var cachePathLocks sync.Map
```
wagoodman: for lib consumers for a long running application this wont really work -- it's an unbounded map that will grow forever

wagoodman: os.Rename is atomic by nature, so we don't really need the sync.Map at all

wagoodman (approved):
I simplified the implementation some:
- removed the sync.map
- more idiomatic cleanup defer function

It should behave the same in terms of atomic swap after the layer contents are ready though.

PR #693, opened 2026-09-16
Fix: read cancellation should wait for workers

wagoodman:
Cancelling an `Image.Read` returned while the layer workers it started were still running.

`async.Collect` treats a done context as permission to give up: it races `ctx.Done()` against its own wait and returns on whichever fires first, so whatever it already handed to the executor keeps going. `readLayers` passed the caller's context straight through to both stages, so a cancelled read handed back layers that fetch and index workers were still writing to. `Read`'s error path then closed those same layers, and the caller's `Cleanup()` deleted the cache directory an indexer was still reading its tar out of.

Anything that cancels a scan hits this: a timeout, a ctrl-C, an HTTP request context. In a consumer's own `-race` suite it surfaces as a race inside stereoscope with no obvious cause. It has been live since the fetch/index overlap landed in #681.

- both stages now run `Collect` on a context that cannot be cancelled (`context.WithoutCancel`), so its return, and `readLayers`', means every worker it started has finished
- cancellation is acted on where that is safe instead: the iterators stop handing out layers, and each stage task returns immediately when the caller's context is already done
- `Image.Read` therefore does not return until in-flight layer work has finished, and now says so along with what the wait costs. `GetImage` and `GetImageFromSource` point at it, and the exported executor names state that a caller-supplied executor has to run every function handed to `Go`

Steady state is untouched: the same work goes to the same executors in the same order. A cancelled read waits out whatever is already in flight. A registry fetch aborts with the caller's context, but a daemon or archive fetch, and any layer index, run to completion. Making those interruptible needs a context threaded through the tar and squashfs walks, which is a separate change.


pkg/image/layer_read_pipeline_test.go
```diff
@@ -701,3 +701,276 @@ func TestImage_Read_squashedSearchContextMatchesAFreshlyBuiltOne(t *testing.T) {
...
+}
+
+func goroutineRunning(frame string) bool {
+	return strings.Contains(allStacks(), frame)
+}
+
+// Read must not hand control back to the caller while goroutines it started are still mutating the
+// layers it is about to release: the error path closes those layers and the caller's Cleanup
+// deletes the cache directory an indexer is still reading its tar out of.
+func TestImage_Read_cancelledReadDoesNotLeaveAnIndexerRunning(t *testing.T) {
```
kzantow, line 803: nit: would be nice if the test functions were at the top

kzantow (approved):

PR #694, opened 2026-09-16
Fix: clear opaque whiteouts by real path, not through symlinks

wagoodman:
An opaque whiteout (`.wh..wh..opq`) in a layer means "hide everything the lower layers put in this one directory". The squash got this wrong in several ways whenever the lower layer had a symlink at that path, deleting a directory the upper layer never mentioned while keeping files the layer did delete. Separately, a malformed link under a regular whiteout failed the whole image read instead of one path.

Each scenario below shows layer contents, then the squashed view before and after.

**1. Deleting through the link.** The opaque directory resolved the path through the lower layer's symlink and cleared the link *target*.

```
layer 1   /target/keep.txt
          /link -> /target

layer 2   /link/  (dir)  /link/.wh..wh..opq  /link/new.txt

today     /link  /link/new.txt  /target                     <- /target/keep.txt deleted

fixed     /link  /link/new.txt  /target  /target/keep.txt
```

An overlay mount stops showing the link and leaves `/target` alone, because `/target` is a separate path that nothing in the upper layer touched. So the squash reported a file as deleted that a running container still exposes. This is the one with real blast radius: a false negative means anything reading the tree silently misses content that is in the image.

**2. Both failures in one image.** A tar can record a file at a path that runs through a symlink, which is how a package installing into `/usr/lib` gets recorded when `/usr/lib` is already a link. Those entries hang off the link itself, and the opaque directory walked straight past them while wiping the target instead.

```
layer 1   /usr/lib -> /usr/lib64
          /usr/lib64/libssl.so.1.1
          /usr/lib/libcrypto.so.1.1        (recorded literally, through the link)

layer 2   /usr/lib/.wh..wh..opq  /usr/lib/libssl.so.3

today     /usr/lib/libssl.so.3  /usr/lib/libcrypto.so.1.1
            ^ libcrypto survived a wipe that covered it
            ^ and /usr/lib64/libssl.so.1.1 was deleted, though nothing asked for that

fixed     /usr/lib/libssl.so.3  /usr/lib64/libssl.so.1.1
```

Both directions at once: a package the image author deleted keeps getting reported, and a package that is really there stops being reported.

**3. Giving up entirely on a link cycle.** When the link could not be followed at all, the resolution error was swallowed and nothing was cleared, on the reasoning that an unresolvable link has no children. It does: they are real nodes addressable by their own real paths.

```
layer 1   /a -> /b   /b -> /a   /a/deleted.txt

layer 2   /a/.wh..wh..opq  /a/kept.txt

today     /a  /b  /a/kept.txt  /a/deleted.txt     <- deleted.txt survived

fixed     /a  /b  /a/kept.txt
```

**4. A regular `.wh.` whiteout over a malformed link killed the read.** The single-file marker this time, not the opaque one. One bad symlink meant no results at all rather than one missing path.

```
layer 1   /b -> /a   /a/gone.txt   /keep.txt

layer 2   /a -> /b   /a/.wh.gone.txt      (upper carries the link, re-forming the cycle)

today     error: filetree Merge failed to remove upperPath (upperPath=/a/gone.txt):
                 cycle during symlink resolution
fixed     /a  /b  /keep.txt
```

Changes:

- an opaque directory now always clears the lower tree's children at the real path, and never follows a link to do it. Scenarios 1, 2 and 3 all collapse into that one rule, and since no link is followed there is nothing left for a malformed one to break
- a malformed link no longer fails the squash in the single-file whiteout branch either
- `RemoveChildPaths` and `RemovePath` keep their contracts, following links and returning the malformed-link error, so the new behavior stays internal to `Merge`


kzantow (approved):
TIL about opaque whiteouts 🤯

PR #695, opened 2026-09-16
fix: resume registry blob downloads dropped mid-stream

juanjsebgarcia:
ENTERPRISE-9338

## Problem

go-containerregistry retries at the request level only (`transport.NewRetry` wraps `RoundTrip`). Once a blob's bytes are flowing, a connection that dies partway through - the registry, a CDN edge, a proxy, the local network - surfaces as an `unexpected EOF`, and the entire layer is discarded however much of it had already arrived:

```
unable to populate layer cache dir="/scratch/.../sha256:ef31d173..." : unexpected EOF
```

Registries fronted by a CDN do this on multi-gigabyte layers - `mcr.microsoft.com` on the Windows base images, where the 4.0 GB first layer of `windows/servercore:ltsc2016` fails repeatedly when the CDN path is degraded. skopeo hits the same drop and recovers by reconnecting, so the blob is fine; only the fetcher gives up.

## Change

`resumableTransport` wraps the RoundTripper so a large `GET` whose body fails is reopened with a `Range` request from the byte already delivered. The gzip and tar readers above it never notice.

**Scope:** a plain `GET`, carrying no body of its own and no caller `Range`, returning `200` in the identity coding with a known `Content-Length` of at least 1 MiB. The length must be known - it is the only way to tell a stream that ended early from one that ended properly, and it keeps transparently decompressed responses out, since `net/http` sets `ContentLength` to `-1` for those. The coding must be identity for the same reason the offsets are wire bytes: an identity segment cannot be spliced onto a stream that arrived in some other coding.

**Corruption is still caught.** This sits *below* go-containerregistry's `verify.ReadCloser`, which tees every byte into the hasher and checks digest and size at EOF - so for a blob a bad splice fails the layer exactly as a corrupt single-shot download would. That backstop belongs to the blob path and not to everything wrapped (a manifest fetched by tag is not digest-verified), so the offset, length and content coding of every resumed response are checked here too, and the body is capped at the length the response declared.

## A safety net, not a heroic recovery

Worth stating plainly, because it is what sizes the budgets. Registry pulls in stereoscope already work well; this is cover for the occasional dropped connection on an otherwise healthy path, not a fix for a pull that is fundamentally failing. A pull needing more than twenty reconnections for one layer is not having a flaky moment, it is on a broken path, and failing it loudly is the right outcome - which is the behaviour stereoscope has today.

Termination is therefore two budgets:

| | |
| --- | --- |
| `maxConsecutiveStalls` = 5 | attempts one drop may take before the read is abandoned |
| `maxTotalResumes` = 20 | attempts for a whole body, however much progress each one makes |

The second is needed because any 64 KiB delivered replenishes the first, so a path dropping every few kilobytes would otherwise reconnect indefinitely while technically advancing - hundreds of thousands of requests on a 4 GB layer, which reads to an operator as a hang rather than a failure. The two move together: the total stays a comfortable multiple of the per-drop one, since a single drop may legitimately spend that whole budget on its own.

Alongside them sits one clock, `reopenHeaderTimeout` = 30 s, bounding how long a *reopen* may take to produce response headers. Without it an endpoint that accepts the connection and then falls silent is ended only by the caller's `context`, which callers do not always bound. It stops the moment headers arrive, so nothing is on a clock during the transfer itself, and the same cancellation lets `Close` abort a reopen that has not yet returned - which an unwrapped `net/http` body does and this otherwise would not.

That puts a number on the worst case. A wholly silent endpoint delivers no progress, so the stall budget ends it after five attempts, under three minutes. The ceiling across every shape is the total budget against the clock - twenty attempts, so about ten minutes - and reaching it needs an endpoint that alternates delivering just enough to replenish the stall budget with falling silent again. Both are pulls that fail outright today.

Explicitly out of scope: keeping a pull alive across an arbitrarily degraded link. No amount of resuming is owed to a connection that cannot hold one.

## Classifying a reopen that is not a 206

Sorted by what another attempt could plausibly mend:

- **`429` and `5xx`** spend a stall and go round again. Treating these as final made a busy CDN indistinguishable from one that has never heard of `Range` - and a single `503` discarded everything already delivered, while a dropped TCP connection from the same host was survivable. The server's own `Retry-After` is deliberately not read: the stall backoff already spaces repeated attempts, and honouring a hint of minutes would only lengthen a pull that is already failing.
- **`401` and `403`** stay permanent. `reopen` reissues the request below go-containerregistry's auth transport, so every attempt carries the same expired token or stale pre-signed URL; retrying turns one honest failure into five.
- **`416`, a plain `200`, a `3xx`, a wrong offset, a changed length** are settled by the server and stay permanent.

On a resumed segment the content-coding check runs *below* the status check, since it guards a body about to be spliced and that is only ever a `206` - running it first made a busy server that serves its error page compressed indistinguishable from one that cannot resume at all. `Content-Encoding: identity` is accepted, because `reopen` asks for identity and some servers echo it straight back.

## Three things learned the hard way

**The range is closed, `bytes=N-M`.** go-containerregistry's *own* registry - what `crane registry serve` and ko run, and what a good many CI harnesses embed - parses the header with `Sscanf("bytes=%d-%d")` and answers **416** to an open-ended `bytes=N-`, which would discard the layer on its first drop and leave things worse than no fix at all. There is an integration test against that registry behind a truncating proxy; restoring the open-ended form makes it fail.

**Resuming is not gated on `Accept-Ranges`.** Probing the largest layer of a real image on each of eight registries - Docker Hub, GHCR, Quay, ECR Public, GCR, registry.k8s.io, GitLab and MCR, the last of these twice so that the 4.0 GB `servercore` layer was covered as well as a small one - all nine blobs honoured a closed range with a `206` at the correct offset, yet six of the eight advertised no `Accept-Ranges` at all. Gating on the header would have disabled resuming on most of them, including the one that motivated this.

**A `3xx` on resume is refused rather than followed.** This transport sits below go-containerregistry's `checkRedirectSSRF`, so following a `Location` from here would reach a host that client never vetted.

## Escape hatch

`RegistryOptions.DisableBlobResume` restores the previous behaviour. Note it is not yet reachable from a released binary: syft, grype and anchorectl would each need to plumb a config key through, which is a follow-up.

## Limitation worth knowing

Every registry probed serves blobs from a **time-limited signed CDN URL** reached by redirect. A transfer that outlives that URL gets a `403` this transport cannot recover from, since minting a fresh one means re-requesting the original registry URL from below the redirect-following client. MCR's window is ~30 minutes and Docker Hub's ~50, so the motivating case may still fail this way on a sufficiently degraded path. Self-hosted registries - Harbor, Artifactory, vanilla `distribution` - serve blobs under the bearer token instead, and `registry:2` issues five-minute tokens by default. Neither is a regression, since such a pull failed before too, but both want the same follow-up: thread a redirect-following, authenticated round-tripper down for reopens.

## Verification

30 tests over the transport plus two integration tests driving the real go-containerregistry stack. They cover reassembly across drops; surviving many drops while making progress; terminating against a trickling server and against one that keeps delivering just enough to replenish the stall budget; the status classification above; refusing a wrong-offset resume, an unusable `Content-Range`, a changed blob length and a genuinely encoded segment; preserving `Authorization` and `User-Agent`; resuming through a 307 to a CDN; never returning more than was declared; HTTP/2, including a `RST_STREAM` after the final byte; read-after-close; `Close` abandoning a backoff and aborting a reopen in flight; cancellation; and leaving small, unknown-length, encoded and bodyless responses unwrapped.

The transport suite passes under `-race`, as do `go vet` and the repo's `closecheck` analyzer. The registry-backed tests in this package need outbound network access and are left to CI.

**Caveat:** the original MCR drop did not reproduce on demand, so the resume path is exercised against fault-injecting test servers and go-containerregistry's own registry rather than a live mid-blob drop.


pkg/image/oci/registry_provider.go
```diff
@@ -257,7 +257,12 @@ func getTransportWithEffectiveURL(tlsConfig *tls.Config, effectiveTransport *eff
...
 	baseTransport := getTransport(tlsConfig)
 
+	// resume large blob downloads that the registry drops mid-stream. this must sit closest to
+	// the base transport so that it wraps the real response body, and below go-containerregistry's
+	// digest verification so that a reassembled layer is still verified end to end
+	resuming := newResumableTransport(baseTransport)
+
 	// wrap it with our effective URL capturing transport
-	effectiveTransport.base = baseTransport
+	effectiveTransport.base = resuming
```
wagoodman: It might be a good idea to make this an opt-out feature through `RegistryOptions`

PR #696, opened 2026-09-16
Fix: whiteout marker correctness

wagoodman:
Two ways a crafted layer entry can wreck a squash, plus whiteout markers leaking out of the lowest layer.

`Path.UnWhiteoutPath` matched the opaque marker with a prefix check while everything around it (`IsDirWhiteout`, the merge walker) matches it exactly. An entry named `.wh..wh..opqX` is therefore not opaque to the walker but still unwhiteouts to its **parent directory**, so a layer carrying `etc/.wh..wh..opqX` erases all of `/etc` from the squash. At the image root it resolves to `/`, which `RemovePath` refuses, and that error aborts the entire read. Per the OCI changeset rules the opaque marker is the file named exactly `.wh..wh..opq`; anything else with a `.wh.` prefix whiteouts the sibling that follows it.

Whiteouts are also only consumed by `FileTree.Merge`, which walks the upper tree. The lowest tree is only copied, so markers in the base layer survived as ordinary files in every squashed tree built on it.

Scenarios below, layers listed bottom first, showing `Image.SquashedTree()` before and after.

**An entry that merely starts with the opaque marker, inside a directory.** `.wh..wh..opqX` means "hide the sibling named `.wh..opqX`", which does not exist, so nothing should happen.

```
L0  etc/passwd, etc/hostname
L1  etc/.wh..wh..opqX

today   (empty: /etc and everything under it is gone)
fixed   /etc, /etc/passwd, /etc/hostname
```

**The same entry at the image root.** It resolves to `/`, which cannot be removed, and the error takes down the whole read rather than one path.

```
L0  etc/passwd
L1  .wh..wh..opqX

today   read fails: cannot remove the root path (`/`) from the FileTree
fixed   /etc, /etc/passwd
```

**Markers in the lowest layer.** Nothing merges into the bottom tree, so its markers were copied through as ordinary files and reported as real content.

```
L0  etc/passwd, etc/.wh.shadow
L1  etc/hostname

today   /etc, /etc/.wh.shadow, /etc/hostname, /etc/passwd
fixed   /etc, /etc/hostname, /etc/passwd
```

**A single-layer image.** Worst case of the same thing: the image squash *was* the bottom tree, so every marker in it showed up as a file.

```
L0  etc/passwd, etc/.wh.shadow

today   /etc, /etc/.wh.shadow, /etc/passwd
fixed   /etc, /etc/passwd
```

**A real opaque marker, unchanged.** The ordinary case still behaves exactly as before.

```
L0  etc/a, etc/b
L1  etc/.wh..wh..opq, etc/c

today   /etc, /etc/c
fixed   /etc, /etc/c
```

- `UnWhiteoutPath` matches the opaque marker exactly, and rejects a path that is not a whiteout instead of returning a best-effort parent for it
- `UnionFileTree.Squash` strips whiteout entries from the copied lowest tree, and now states the contract it had been leaving implicit: a squash never carries markers, so squashing a single tree is not the same as copying it
- the image squash no longer aliases layer 0's diff tree as its squashed tree, it squashes it like every other layer. Diff trees keep their own markers, which is where a merge reads them from
- layer 0's squash now reports progress, which the read progress total (`layers * 2`) already assumed

One tradeoff: layer 0's squashed tree used to alias its diff tree, so it was free. It is now a real copy retained for the life of the image, on what is usually the largest layer.


wagoodman:
The unit test failure is being addressed in #693 ...

kzantow (approved):

